PR #1, opened 2016-06-10
Initial move from polymer-cli

justinfagnani:
Only minimal changes to make tests and lint pass.


justinfagnani:
Yes, I did the minimal move so that this is rubber stamped, and subsequent changes are actually reviewable. I only changed some paths, and trimmed down the dependencies.

PR #2, opened 2016-06-13
HtmlProject -> PolymerProject and add build steps. Add example project.

justinfagnani:


justinfagnani:
cc @robdodson

test/test-project/gulpfile.js
```diff
@@ -0,0 +1,40 @@
...
+    .pipe(project.analyze);
+
+  let bundled = fork(allFiles)
+    .pipe(project.bundle)
+    .pipe(gulp.dest('build/bundled'));
+
+  let unbundled = fork(allFiles)
+    .pipe(gulp.dest('build/unbundled'));
+
+  return allFiles;
```
rictic: Recommend returning some combination of bundled and unbundled here

src/fork-stream.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Forks a stream of Vinyl files, cloning each file before emitting on the fork.
+ */
+export class ForkedVinylStream extends Readable {
+
+  input: NodeJS.ReadableStream;
+
+  constructor(input: NodeJS.ReadableStream) {
+    super({objectMode: true});
+    this.input = input;
```
rictic: The three declaration/assignments of `input` here are one reason why I like just putting `public` on the constructor argument.

YMMV.

src/fork-stream.ts
```diff
@@ -0,0 +1,42 @@
...
+import {PassThrough, Readable, Transform} from 'stream';
+
+export function forkStream(stream: NodeJS.ReadableStream) {
+  return new ForkedVinylStream(stream);
+}
+
+/**
+ * Forks a stream of Vinyl files, cloning each file before emitting on the fork.
+ */
+export class ForkedVinylStream extends Readable {
```
rictic: Did you consider extending stream.Transform instead? That way most of the piping of input to output is handled automagically. It might be as simple as:

```javascript

import * as stream from 'stream';

export function forkStream(input: NodeJS.ReadableStream): NodeJS.WritableStream {
  const forkedStream = new ForkedVinylStream();
  forkedStream.pipe(input);
  return forkedStream;
}

class ForkedVinylStream extends stream.Transform {
  constructor() {
    super({objectMode: true});
  }
  _transform(file) {
    this.push(file.clone({deep: true, contents: true}));
  }
}
```

src/polymer-project.ts
```diff
@@ -26,25 +31,186 @@ const extensionsForType = {
...
+  entrypoint?: string;
+
+  /**
+   * The app shell element.
+   */
+  shell?: string;
+
+  /**
+   * Lazily loaded fragments. Usually the pages of an app or other bundles of
+   * on-demand resources.
```
rictic: Are these paths to html files on disk?

src/polymer-project.ts
```diff
@@ -26,25 +31,186 @@ const extensionsForType = {
...
+   */
+  fragments?: string[];
+
+  /**
+   * List of glob patterns to read from the file system.
+   */
+  sourceGlobs?: string[];
+
+  /**
+   * List of glob patterns to read from the file system.
```
rictic: What's a dependency in this context?

src/polymer-project.ts
```diff
@@ -26,25 +31,186 @@ const extensionsForType = {
...
+   */
+  dependencyGlobs?: string[];
+}
+
+export const defaultSourceGlobs = [
+  '**/*',
+  // '!build',
```
rictic: Remove comment?

src/polymer-project.ts
```diff
@@ -26,25 +31,186 @@ const extensionsForType = {
...
+    });
+  }
+
+  // TODO(justinfagnani): add allFiles()
+
+  split(): Transform {
+    return new Splitter(this);
+  }
+
+  rejoin(): Transform {
```
rictic: Why a mix of getters and zero-arg methods?

src/project-config.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  static fromConfigFile(filepath: string): ProjectConfigOptions {
+    try {
+      let configContent = fs.readFileSync(filepath, 'utf-8');
+      return JSON.parse(configContent);
+    } catch (error) {
+      if (error.code === 'ENOENT') {
+        logger.debug('config file not found', {path: filepath});
+      } else {
+        logger.warn('Could not load config file', {path: filepath, err: error});
```
rictic: Should this be an error?

src/project-config.ts
```diff
@@ -0,0 +1,88 @@
...
+      }
+      return {};
+    }
+  }
+
+  constructor(defaultOptions?: string | ProjectConfigOptions, overrideOptions?: ProjectConfigOptions) {
+    if (typeof defaultOptions === 'string') {
+      defaultOptions = ProjectConfig.fromConfigFile(<string>defaultOptions);
```
rictic: This cast is going away in typescript >1.8

src/project-config.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export interface ProjectConfigOptions {
+  root?: string;
+  entrypoint?: string;
+  shell?: string;
+  fragment?: string[];
+  fragments?: string[];
+}
+
+export class ProjectConfig {
```
rictic: I count two interface and one class with very very similar sets of properties.

ProjectConfigOptions, ProjectConfig, and ProjectOptions. How many of these are exposed to users? Can some of them implement / extend others?

test/test-project/gulpfile.js
```diff
@@ -0,0 +1,52 @@
...
+    // TODO(justinfagnani): allow filtering of files before writing
+    .pipe(gulp.dest('build/bundled'));
+
+  let unbundled = fork(allFiles)
+    // write to the unbundled folder
+    // TODO(justinfagnani): allow filtering of files before writing
+    .pipe(gulp.dest('build/unbundled'));
+
+  return mergeStream(bundled, unbundled);
+});
```
rictic: Don't forget a couple of tests! Just something super simple to start with that runs gulp on the test project and does a byte for byte comparison of the output to known good output would be awesome.

src/fork-stream.ts
justinfagnani: I just don't like more than one way to do things so fundamental... I want to look in one place for the class fields. I could make a local var, but, eh...

src/fork-stream.ts
justinfagnani: Let me try that

justinfagnani: Worked, but needed `callback(null, file)` instead of `push`. Thanks!

src/polymer-project.ts
justinfagnani: yes, and relative to root. Clarifying.

src/polymer-project.ts
justinfagnani: done

src/polymer-project.ts
justinfagnani: done

src/polymer-project.ts
justinfagnani: I tried to make them all nullary methods, but for the analyzer and bundler they needed to be specific instances because they reference each other. Splitter and joiner on the other hand need to be separate instances because they appear in multiple forks of the build pipeline.

src/project-config.ts
justinfagnani: Maybe? I'm actually not using this yet... I could remove the file

src/project-config.ts
justinfagnani: The difference is in the optionality and in fragment/fragments. One is produced by args parsing, the other passed to the API. I made a new one here optimized for calling from gulp... I'll remove this file for now and reconcile later. Sorry for including this.

test/test-project/gulpfile.js
justinfagnani: oh yeah... forgot to move the tests around. See polymer-project_test.js with tests sources, dependencies, split and join. Bundling is already tested outside of the project class. I'll keep adding more.

justinfagnani:
PTAL

src/polymer-project.ts
rictic: Seems like a subtle but important constraint. Mind documenting it here?

src/fork-stream.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Forks a stream of Vinyl files, cloning each file before emitting on the fork.
+ */
+export class ForkedVinylStream extends Transform {
+
+  constructor() {
+    super({objectMode: true});
+  }
+
+  _transform(file: File, enc: string, callback) {
```
rictic: No type annotation for callback? Does this warn under no-implicit-any, or is picked up by the fact that you're overriding the Transform method?

src/fork-stream.ts
```diff
@@ -0,0 +1,32 @@
...
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import {PassThrough, Readable, Transform} from 'stream';
+import File = require('vinyl');
+
+export function forkStream(stream: NodeJS.ReadableStream) {
```
rictic: Consider annotating the return value as a NodeJS.ReadableStream because it's not intended that the receiver write to it.

rictic:
Looks like tests are failing from there not being an `init` script in `package.json`.

src/fork-stream.ts
justinfagnani: I think picked up by the override, but also good point: I enabled `noImplicityAny` and disabled `suppressImplicitAnyIndexErrors`. Still no errors.

justinfagnani: Whoops... `gulp build` was showing no errors, but `tsc` is. Looking into the `gulp` issue...

src/fork-stream.ts
justinfagnani: done

justinfagnani:
PTAL

Fixed all the things. `gulp build` reports type errors, the project lints, added npm scripts, so it should pass on CI, etc..

gulpfile.js
```diff
@@ -18,21 +18,37 @@ const mergeStream = require('merge-stream');
...
 const runSeq = require('run-sequence');
-const tslint = require("gulp-tslint");
+const stream = require('stream');
+const tslint = require('gulp-tslint');
 const typescript = require('gulp-typescript');
 const typings = require('gulp-typings');
 
 const tsProject = typescript.createProject('tsconfig.json');
 
+// WTF gulp-typescript
```
rictic: ?

gulpfile.js
```diff
@@ -18,21 +18,37 @@ const mergeStream = require('merge-stream');
...
+  _transform(file, enc, callback) {
+    let oldPath = file.path;
+    if (oldPath.startsWith('lib/src')) {
+      file.path = 'lib' + oldPath.substring('lib/src'.length);
+    }
+    callback(null, file);
+  }
+}
+
 gulp.task('init', () => gulp.src("./typings.json").pipe(typings()));
```
rictic: Looks like this isn't working now? typings aren't found in travis

rictic:
Btw, now that github supports squash and merge I think we should settle on pushing incremental commits up for review. Makes it way easier for me to review new changes.

justinfagnani:
Oops, sorry about the squashing. I still prefer merge's because I often rebase into a few logical commits.

justinfagnani:
Ok, updated gulp-typings which uses a newer version of typings which writes index.d.ts files instead of main/browser.d.ts so init should work...

rictic:
LGTM

PR #4, opened 2016-06-30
Create a smaller, smarter dependency tree during build

FredKSchott:
*Note: Tests still in progress, but the code itself has been cleaned up and can be considered ready for review. I'm submitting now to get some early feedback.*

Known limitations:
- [x] Source files can be dependencies of other source files, so those that are will currently come out of both streams. Should they be filtered out of the dependency stream?  ANSWER: Yes!
- [x] Users can only specify single files to be included on top of the dependency tree. Should we add support for globs? Should we add support for including their transitive dependencies as well? ANSWER: Support for globs added, transitive dependency detection can be a future enhancement if enough people ask for it.

/cc @justinfagnani for feedback on limitations and if they need to be prioritized? I think we could release right now without both but I'm not 100% confident in what PSK does and doesn't need.

### DIFF
--- a/src/document-parser.ts
+++ b/src/document-parser.ts
@@ -1,0 +1,71 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import {DocumentDescriptor} from 'hydrolysis';
+import {posix as posixPath} from 'path';
+import {Node, queryAll, predicates, getAttribute} from 'dom5';
+
+export interface DocumentDeps {
+  imports?: Array<string>;
+  scripts?: Array<string>;
+  styles?: Array<string>;
+}
+
+function collectScriptsAndStyles(tree: DocumentDescriptor): DocumentDeps {
+  let scripts: string[] = [];
+  let styles: string[] = [];
+  tree.html.script.forEach((script: Node) => {
+    // TODO(justinfagnani): stop patching Nodes in Hydrolysis
+    let __hydrolysisInlined = (<any>script).__hydrolysisInlined;
+    if (__hydrolysisInlined) {
+      scripts.push(__hydrolysisInlined);
+    }
+  });
+  tree.html.style.forEach((style: Node) => {
+    let href = getAttribute(style, 'href');
+    if (href) {
+      styles.push(href);
+    }
+  });
+  return {
+    scripts,
+    styles
+  };
+}
+
+export function getDependenciesFromDocument(descriptor: DocumentDescriptor, dir: string): DocumentDeps {
+  let allHtmlDeps: string[] = [];
+  let allScriptDeps = new Set<string>();
+  let allStyleDeps = new Set<string>();
+
+  let deps: DocumentDeps = collectScriptsAndStyles(descriptor);
+  deps.scripts.forEach((s) => allScriptDeps.add(posixPath.resolve(dir, s)));
+  deps.styles.forEach((s) => allStyleDeps.add(posixPath.resolve(dir, s)));
+  if (descriptor.imports) {
+    let queue = descriptor.imports.slice();
+    while (queue.length > 0) {
+      let next = queue.shift();
+      if (!next.href) {
+        continue;
+      }
+      allHtmlDeps.push(next.href);
+      let childDeps = getDependenciesFromDocument(next, posixPath.dirname(next.href));
+      allHtmlDeps = allHtmlDeps.concat(childDeps.imports);
+      childDeps.scripts.forEach((s) => allScriptDeps.add(s));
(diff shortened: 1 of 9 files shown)


justinfagnani:
@FredKSchott Have you looked at the CI failures?

justinfagnani:
re: limitations. It's very unlikely that third party things depend on sources. That would be kind of crazy actually. But it could happen, and filtering would be appropriate.

We should allow globs for including third party files though.

FredKSchott:
@justinfagnani the more common scenario is `src/source-a.html` depending on `src/source-b.html`. Our current pass through hydrolysis will flag `src/source-b.html` as a dependency, which it technically is, even though it is also a source file.

But the more I think about this, the more obvious it is that `source-b.html` should not be included in the dependency stream because it is already a source. I'll make that change (no sources in dependency stream) and then any more advanced filtering can be a "future work" enhancement.

FredKSchott:
Enhancements discussed in PR body have been addressed, starting on test suite now

src/document-parser.ts
```diff
@@ -0,0 +1,61 @@
+import {DocumentDescriptor} from 'hydrolysis';
```
justinfagnani: copyright notice

FredKSchott: This was grabbed out of analyzer (see above). Was it originally from some external source?

justinfagnani: no, we just put a copyright notice at the top of all files

src/polymer-project.ts
```diff
@@ -62,20 +62,17 @@ export interface ProjectOptions {
...
-  '**/*',
-  '!build/**/*',
-];
-
-export const defaultDependencyGlobs = [
-  'bower_components/**/*',
-  'node_modules/**/*',
+  'src/**/*',
+  // NOTE(fks) 06-29-2016: `polymer-cli serve` uses a bower.json file to display
+  // information about the project. The file is included here by default.
```
justinfagnani: It'd be nice to not require this anymore. The file is used only to grab a package name when serving component-style

src/polymer-project.ts
```diff
@@ -62,20 +62,17 @@ export interface ProjectOptions {
   sourceGlobs?: string[];
 
   /**
-   * List of glob patterns, relative to root, of dependencies to read from the
-   * file system. For example node_modules\/**\/* and bower_components\/**\/*
+   * List of file paths, relative to the project directory, that should be included
+   * as dependencies in the build target.
    */
-  dependencyGlobs?: string[];
+  includeDependencies?: string[];
```
justinfagnani: I've been thinking we want a new term instead of 'dependencies'. Maybe 'third party'?

src/document-parser.ts
FredKSchott: oh doy, good catch

src/streams.ts
```diff
@@ -42,3 +55,116 @@ export function compose(streams: NodeJS.ReadWriteStream[]) {
     return new PassThrough({objectMode: true});
   }
 }
+
+/**
+ * A stream that accepts source file objects from a vinyl stream, and emits
+ * the file paths of all required file dependencies. Useful for building
+ * the dependency tree during build.
+ */
+export class SourceDependenciesStream extends Transform {
```
justinfagnani: I think this deserves to be in it's own file, or bundled in analyzer.ts

src/streams.ts
```diff
@@ -42,3 +55,116 @@ export function compose(streams: NodeJS.ReadWriteStream[]) {
...
+
+  /**
+   * Create an analyzer and retreive all transitive dependencies for `url`
+   */
+  _getDependencies(url: string): Promise<DocumentDeps> {
+    let pathDirectory = posixPath.dirname(url);
+    let resolver = new FSResolver({});
+    let loader = new Loader();
+    loader.addResolver(resolver);
+    let analyzer = new Analyzer(false, loader);
```
justinfagnani: We already have an analyzer in the project, can't you just use that? Also, seems expensive to new up an Analyzer for each file.

src/streams.ts
```diff
@@ -42,3 +55,116 @@ export function compose(streams: NodeJS.ReadWriteStream[]) {
...
+
+    return analyzer.metadataTree(url)
+      .then((tree) => getDependenciesFromDocument(tree, pathDirectory));
+  }
+}
+
+/**
+ * A stream that passes through Vinyl file objects and loads any that have
```
justinfagnani: when is this happening?

src/document-parser.ts
FredKSchott: done

src/polymer-project.ts
```diff
@@ -181,18 +173,18 @@ export class PolymerProject {
...
 
-  // TODO(justinfagnani): add options, pass to vfs.src()
   dependencies(): NodeJS.ReadableStream {
-    let deps = this.dependencyGlobs;
-    return vfs.src(deps, {
-      allowEmpty: true,
-      cwdbase: true,
-      nodir: true,
-    });
+    return this.sources()
```
justinfagnani: I think this could be something like `return this._analyzer.dependencyPaths()`, where dependencyPaths() is fed files from `StreamResolver.accept()`.

src/polymer-project.ts
FredKSchott: I know I agree it seems unnecessary. It would be nice if it was optional, and if the bower.json file wasn't present it would just not display that extra component-style information. 

I'll create a quick PR on polymer-serve to make it optional, but lets make that change in a future PR / not have that block this.

FredKSchott:
Closing in favor of #5

PR #5, opened 2016-07-14
adds smarter dependency stream via new analyzer architecture

### LINKED ISSUE #3: [Feature] Smarter, smaller builds
## TODO:
- [ ] 1. Stream starts with src/ directory instead of all files, can be changed via user-input globs
- [ ] 2. Dependencies stream involves reading the source stream and parsing the dependency tree via hydrolosis
- [ ] 3. If dependency is found, file is pushed into the top of the build pipeline stream

#### For 3: 
This requires that all files are compiled and present at the start of build pipeline, or that there's a fallback process if file does not exist (assumes it will exist in the future). Possible implementations:
- Force "compilation" step to complete before the build pipeline starts. Ex: run typescript and start build pipeline after all files are created. If we can rely on this assumption, then we can throw an error when a file is not found. Less moving pieces, but also less efficient.
- Handle not found files by putting them into a queue and retrying later. Once all pipeline steps before read are complete, retry the files in this queue. Throw an error if they aren't found at this point. This allows for compilation of all non-html files, but would fail if the imports/loading graph changes during "compilation". More efficient, but puts the restriction on HTML files that they cannot be modified during "compilation". 

@justinfagnani thoughts on this tradeoff? Are we okay with allowing some files to be changed during "compilation" but breaking on certain types of changes (related to HTML files and the imports/scripts/styles loading graph it creates). I'd rather encourage an explicit "before build" step that prepares the target for polymer-build vs. modifying files during read.

FredKSchott:
~~*Note: Some JSDoc and a few more tests are still TODO, just wanted to get this out there today*~~

This PR adds a much smarter dependency stream based on dependency analysis of the application rather than simple glob matching. When run on some of our basic templates, build directories are something like 90+% smaller than they were before.

See `/test/test-project/gulpfile.js` for usage examples.

/cc @justinfagnani @robdodson
Resolves #3

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -9,25 +9,31 @@
  */
 
 import * as fs from 'fs';
-import {Analyzer, Deferred, Loader, Resolver, DocumentDescriptor} from 'hydrolysis';
-import {posix as posixPath} from 'path';
-import * as osPath from 'path';
-import {Transform} from 'stream';
+import {Analyzer, Deferred, Loader, Resolver, DocumentDescriptor}
+  from 'hydrolysis';
+import * as path from 'path';
+import {PassThrough, Transform} from 'stream';
 import File = require('vinyl');
 import {parse as parseUrl} from 'url';
 import * as logging from 'plylog';
 import {Node, queryAll, predicates, getAttribute} from 'dom5';
 
-import {FileCB} from './streams';
-import urlFromPath from './url-from-path';
+import {FileCB, VinylReaderTransform} from './streams';
+import {urlFromPath, pathFromUrl} from './path-transformers';
+import {DocumentDeps, getDependenciesFromDocument}
+  from './get-dependencies-from-document';
 
 const minimatchAll = require('minimatch-all');
 const logger = logging.getLogger('cli.build.analyzer');
 
-export interface DocumentDeps {
-  imports?: Array<string>;
-  scripts?: Array<string>;
-  styles?: Array<string>;
+export interface DepsIndex {
+  // An index of dependency -> fragments that depend on it
+  depsToFragments: Map<string, string[]>;
+  // TODO(garlicnation): Remove this map.
+  // An index of fragments -> html dependencies
+  fragmentToDeps: Map<string, string[]>;
+  // A map from frament urls to html, js, and css dependencies.
+  fragmentToFullDeps: Map<string, DocumentDeps>;
 }
 
 export class StreamAnalyzer extends Transform {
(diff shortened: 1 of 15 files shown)


FredKSchott:
@justinfagnani tests added, PTAL

src/analyzer.ts
```diff
@@ -115,158 +176,94 @@ export class StreamAnalyzer extends Transform {
...
+
+  hasDeferredFiles(): boolean {
+    return this.deferredFiles.size > 0;
+  }
+
+  deferLoading(url: string, deferred: Deferred<string>): void {
+    this.deferredFiles.set(url, deferred);
+  }
+
+  loadDeferredFile(url: string, file: File): void {
```
justinfagnani: naming nit: "load" sounds like it'll cause it to load. maybe `resolveDeferredFile`?

src/analyzer.ts
```diff
@@ -115,158 +176,94 @@ export class StreamAnalyzer extends Transform {
...
 
+  hasDeferredFile(url: string): boolean {
+    return this.deferredFiles.has(url);
+  }
+
+  hasDeferredFiles(): boolean {
+    return this.deferredFiles.size > 0;
+  }
+
+  deferLoading(url: string, deferred: Deferred<string>): void {
```
justinfagnani: this function is only called in one place, internal to this class. probably unnecessary

src/document-parser.ts
```diff
@@ -0,0 +1,71 @@
+/**
```
justinfagnani: The name makes me thing theres a parser in here. `document-utils`?

justinfagnani: Or `getDependenciesFromDocument` in kebab-case

src/polymer-project.ts
```diff
@@ -181,17 +175,25 @@ export class PolymerProject {
...
-  // TODO(justinfagnani): add options, pass to vfs.src()
   dependencies(): NodeJS.ReadableStream {
-    let deps = this.dependencyGlobs;
-    return vfs.src(deps, {
-      allowEmpty: true,
-      cwdbase: true,
-      nodir: true,
-    });
-  }
+    let dependenciesStream: NodeJS.ReadableStream = forkStream(
```
justinfagnani: should this be set-up one time in the constructor?

src/streams.ts
```diff
@@ -42,3 +43,36 @@ export function compose(streams: NodeJS.ReadWriteStream[]) {
     return new PassThrough({objectMode: true});
   }
 }
+
+/**
+ * A stream that passes through Vinyl file objects and loads any files
+ * that have not yet had their content loaded.
```
justinfagnani: when is this needed?

src/analyzer.ts
```diff
@@ -37,33 +41,44 @@ export class StreamAnalyzer extends Transform {
...
+
+  analyzeDependencies: Promise<DepsIndex>;
+  _dependencyAnalysis: DepsIndex = {
+    depsToFragments: new Map(),
+    fragmentToDeps: new Map(),
+    fragmentToFullDeps: new Map()
+  };
+  _resolveDependencyAnalysis: (index: DepsIndex) => void;
+
+  constructor(root: string, entrypoint: string, shell: string, fragments: string[], sourceGlobs: string[]) {
```
justinfagnani: long line?

src/streams.ts
```diff
@@ -42,3 +43,36 @@ export function compose(streams: NodeJS.ReadWriteStream[]) {
...
+    encoding: string,
+    callback: (error?: Error, data?: File) => void
+  ): void {
+    // Don't load a file that already has contents.
+    if (file.contents) {
+      callback(null, file);
+      return;
+    }
+
+    fs.readFile(file.path, (err?: Error, data?: Buffer) => {
```
justinfagnani: can we just move this into SteamAnalyzer.pushDependency()?

src/streams.ts
justinfagnani: after our talk, turn this into a VinylReaderTransform and make the StreamAnalyzer's dependencies stream a stream of filenames.

src/analyzer.ts
FredKSchott: good call, done

src/analyzer.ts
FredKSchott: good call, done

src/analyzer.ts
FredKSchott: whoops! done

src/streams.ts
FredKSchott: As discussed in person, we rely on this stream to coordinate asynchronous file loading that doesn't block or interfere with the analyzer stream.

src/streams.ts
FredKSchott: done!

src/document-parser.ts
FredKSchott: Yea, I guess parser is already a used term in our tooling/build ecosystem.
Changing to `get-dependencies-from-document.ts`

src/polymer-project.ts
FredKSchott: I built this to purposefully match `.sources()`, which currently creates a brand new readable streams for the consumer. 

This sounds like a somewhat useful feature (you could create additional streams for additional build tasks without worrying about how data is modified) but you could do the same with fork() on an as needed basis.

I'm good with either as long as both `sources()` & `dependencies()` behave the same. I'll keep it as is since I don't see a good reason for/against but let me know if I'm missing something.

FredKSchott:
@justinfagnani PTAL

justinfagnani:
LGTM

PR #6, opened 2016-07-15
Add service-worker generation

FredKSchott:
***Note: Some JSDoc and a few more tests are still TODO, just wanted to get this out there today. ~~Also, this relies on the `dependency-stream` work, which is still pending CR. DO NOT MERGE until that has been first merged.~~***

It is now possible to add a service worker via the polymer-build library. Ex:

```
polymerProject.addServiceWorker('build/unbundled', {swConfig: swConfig});
```

This replaces the current code that was build specifically for the polymer-cli use-case, and allows custom configuration via the options object. It is called after the build is finished, since the process relies on the exact files and their file information.

See https://github.com/Polymer/polymer-cli/commit/4c4f36aff651131e121635737292771108617eb7 for it's usage in polymer-cli.

/cc @justinfagnani @robdodson


justinfagnani:
@FredKSchott about to check this out now.

One quick thought about the API beforehand though. I wonder if we'd be better off making it possible/easy for something like this to be written externally to the PolymerProject class. Maybe usage should look like this:

```javascript
  let unbundled = fork(buildStream)
    .pipe(addServiceWorker(project, 'build-5/unbundled', {config: swConfig}))
    .pipe(gulp.dest('build-5/unbundled'));
```

FredKSchott:
@justinfagnani PTAL. 

As discussed, service worker generation actually couldn't exist in the build stream without some nontrivial refactoring, due to the internal sw-precache library reading information directly from the file system. This has replaced the stream interface with a Promise-based one, which is called after generation and resolves once the service worker has been actually written to the provided build directory.

FredKSchott:
As for pulling service worker generation out of `PolymerProject` and into its own function, do you mean to still keep it as a function inside of `polymer-build` or as a separate library? I don't have a preference if its staying in this library, especially after the most recent change away from the stream interface. It feels too early to refactor this into its own library tho, since it adds maintenance costs as you mentioned. I'd rather discuss that kind of refactoring down the road after some real world usage, vs. adding the maintenance burden preemptively.

justinfagnani:
we could keep it in polymer-build package as a sub-library.

FredKSchott:
99% of the logic is around generating the proper `staticFileGlobs` based on the project. What if we went the other way and provided a function like `getSWFilesFromProject()` that just returned that `staticFileGlobs` list, and then the consumer would pass it to the `sw-precache` library `generate()`/`write()` function of choice themselves. That would slim down our responsibilities and dependency graph and give the consumer a little more control over the process. But adding that extra step may discourage some from generating a service worker, so if one of our goals is overall service-worker adoption then this could get in the way.

ex as a service worker generator (current):
```js
return waitForAll([
  bundled, 
  unbundled,
]).then(() => {
  return Promise.all([
    polymerProject.addServiceWorker('build/unbundled', {
      swConfig: swConfig,
      serviceWorkerPath: 'test-custom-sw-path.js',
    }),
    polymerProject.addServiceWorker('build/bundled', {
      swConfig: swConfig,
      bundled: true,
    }),
  ]);
});
```

ex as a file list generator (proposed):
```js
const swPrecache = require('sw-precache');
return waitForAll([
  getSWFilesFromProject(polymerProject),
  getBundledSWFilesFromProject(polymerProject),
  bundled,
  unbundled,
]).then((results) => {
  let unbundledSWFilePaths = results[0];
  let bundledSWFilePaths = results[1];
  return Promise.all([
    swPrecache.write('build/unbundled/test-custom-sw-path.js', {
      staticFileGlobs: unbundledSWFilePaths
    }),
    swPrecache.write('build/bundled/service-worker.js', {
      staticFileGlobs: bundledSWFilePaths
    }),
  ]);
});
```

@justinfagnani what do you think?

FredKSchott:
Ah actually you know what, I forgot we use the buildRoot to generate that list of files. If we have to start passing two different options objects at two different times this becomes a lot harder to swallow. Lets stick with the standalone function for now.

FredKSchott:
Superseded by #7

PR #7, opened 2016-07-26
Add ability to generate and write a service worker to disk

FredKSchott:
It is now possible to add a service worker via the polymer-build library. Ex:

```js
export {addServiceWorker} from 'polymer-build';

addServiceWorker({
  project: polymerProject,
  buildRoot: 'build/unbundled',
  swConfig: swConfig
});
```

This replaces the current code that was build specifically for the polymer-cli use-case, and allows custom configuration via the options object. It must be called after the build is finished, since the internal process relies on the exact files to exist on disk.

See test/gulpfile.js for usage.

/cc @justinfagnani @robdodson


robdodson:
does the current polymer-build not do this step for you?

filipproch:
Hi there, just a minor issue causing test-project/gulpfile.js not to work
```
let allFiles = mergeStream(sources, dependencies)
     .pipe(project.analyze);
 
   // fork the stream in case downstream transformers mutate the files
   // this fork will vulcanize the project
   let bundled = fork(allFiles)
     .pipe(project.bundle)
```

should be also updated to

```
let allFiles = mergeStream(sources, dependencies)
    .pipe(project.analyzer)

  // fork the stream in case downstream transformers mutate the files
  // this fork will vulcanize the project
  let bundled = fork(allFiles)
    .pipe(project.bundler)
    // write to the bundled folder
    .pipe(gulp.dest('build/bundled'))
```

FredKSchott:
@robdodson `polymer-cli` does currently do this for you, `polymer-build` does not.
@filipproch Thanks! good catch

src/polymer-project.ts
```diff
@@ -107,8 +105,20 @@ export class PolymerProject {
 
   _splitFiles: Map<string, SplitFile> = new Map();
   _parts: Map<string, SplitFile> = new Map();
-  _analyzer: StreamAnalyzer;
-  _bundler: Bundler;
+
+  /**
+   * analyzer is a `Transform` that runs Hydrolysis analysis on the files, for
+   * use by the bundler transform. This transform must only be used in one
```
justinfagnani: "bundler transform or any other build step that needs dependency information"?

src/polymer-project.ts
```diff
@@ -107,8 +105,20 @@ export class PolymerProject {
 
   _splitFiles: Map<string, SplitFile> = new Map();
   _parts: Map<string, SplitFile> = new Map();
-  _analyzer: StreamAnalyzer;
-  _bundler: Bundler;
+
+  /**
+   * analyzer is a `Transform` that runs Hydrolysis analysis on the files, for
```
justinfagnani: You don't need the "analyze is" part. Just: "A `Transform`...". Same with bundler

src/polymer-project.ts
```diff
@@ -107,8 +105,20 @@ export class PolymerProject {
...
+   * use by the bundler transform. This transform must only be used in one
+   * stream.
+   */
+  analyzer: StreamAnalyzer;
+
+  /**
+   * bundler is a `Transform` that bundles the shell and fragments of the
```
justinfagnani: This doesn't really describe what bundling actually is :)

src/polymer-project.ts
```diff
@@ -214,26 +224,6 @@ export class PolymerProject {
     return new HtmlRejoiner(this);
   }
 
-  /**
-   * Returns the a `Transform` that runs Hydrolysis analysis on the files, for
-   * use by the bundler transform. This transform must only be used in one
-   * stream.
-   */
-  get analyze(): Transform {
```
justinfagnani: Why'd you remove the getters?

src/service-worker.ts
```diff
@@ -0,0 +1,153 @@
...
+ */
+
+import {writeFile} from 'fs';
+import * as path from 'path';
+import * as logging from 'plylog';
+import {PolymerProject} from './polymer-project';
+import {DepsIndex} from './analyzer';
+const swPrecache = require('sw-precache');
+
+// non-ES compatible modules
```
justinfagnani: this isn't a module

src/service-worker.ts
```diff
@@ -0,0 +1,153 @@
...
+let logger = logging.getLogger('polymer-build.service-worker');
+
+export interface SWConfig {
```
justinfagnani: Is this interface just describing an interface of the sw-precache library? If so, it should live in custom_typings, and they you can import sw-precache as a module

src/service-worker.ts
```diff
@@ -0,0 +1,153 @@
...
+  staticFileGlobs?: string[];
+  stripPrefix?: string;
+  templateFilePath?: string;
+  verbose?: boolean;
+}
+
+export interface AddServiceWorkerOptions {
+  project: PolymerProject;
+  buildRoot: string;
+  bundled?: boolean;
```
justinfagnani: it would be nice if this wasn't necessary...

src/service-worker.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+/**
+ * Returns a service worker transform stream. This stream will add a service
+ * worker to the build stream, based on the options passed and build analysis
+ * performed eariler in the stream.
+ *
+ * Note that this stream closely resembles a pass-through stream. It does not
+ * modify the files that pass through it. It only ever adds 1 file.
+ */
```
justinfagnani: It might be nice to have example usage in the comment

src/service-worker.ts
```diff
@@ -0,0 +1,153 @@
...
+  return Array.from(precachedAssets).concat(project.bundler.sharedBundleUrl);
+}
+
+/**
+ * Returns a service worker transform stream. This stream will add a service
+ * worker to the build stream, based on the options passed and build analysis
+ * performed eariler in the stream.
+ *
+ * Note that this stream closely resembles a pass-through stream. It does not
+ * modify the files that pass through it. It only ever adds 1 file.
```
justinfagnani: Can you name the file it adds?

src/service-worker.ts
```diff
@@ -0,0 +1,153 @@
...
+ */
+export function generateServiceWorker(options: AddServiceWorkerOptions): Promise<Buffer> {
+  console.assert(!!options.project, '`project` option is required');
```
justinfagnani: I prefer `options.project != null` for being more explicit and intent preserving, but then again I'm a little weird...

src/polymer-project.ts
FredKSchott: It seemed like these were built when the analyzer/bundler were only used as streams. Now that the analyzer is a part of the public interface (service-worker generation calls `analyzer.analyzeDependencies()`) the getter/alias becomes unnecessary and the `-er` form became more fitting.

src/service-worker.ts
FredKSchott: Good call!

src/service-worker.ts
FredKSchott: +1 point for the build stream as its own custom type :)

We could split the two methods and add `addBundledServiceWorker()`. `generateServiceWorker()` could remain as-is so that we don't need to split that as well.

We could also make `getPrecachedAssets()`/`getBundledPrecachedAssets()` public and have the gulpfile author properly call those themselves, and then send the result as an option to `generateServiceWorker()` instead.

I don't love it either but it's a small nit for me vs. our other options. Let me know if you think this is a big enough problem that we should address it now.

src/service-worker.ts
FredKSchott: isn't that less explicit? Because of the type conversion it's also matching all falsey values, even though it looks like its just checking for null. The code as written is explicitly casting it to boolean before checking falsey-ness.

FredKSchott:
@justinfagnani Thanks for the feedback on documentation! Going back over all docs and making sure they're still up-to-date now

src/polymer-project.ts
FredKSchott: Bundler(n): a thing that bundles

FredKSchott:
@justinfagnani PTAL

*Note: Tests are failing on travis because of that unrelated issue in one of our dependencies (gulp-typings -> beautylog) that I mentioned this morning. This is only an issue for development and does not affect library usage.*

*There's a PR in to get this fixed but if it's really bothering us we can drop versions until it's settled.*

src/service-worker.ts
justinfagnani: No, this is fine for now. Let's just ruminate...

src/service-worker.ts
```diff
@@ -0,0 +1,125 @@
...
+ * the project's required service worker precached assets.
+ */
+function getPrecachedAssets(depsIndex: DepsIndex, project: PolymerProject): string[] {
+  let precachedAssets = new Set<string>(project.analyzer.allFragments);
+  precachedAssets.add(project.entrypoint);
+
+  for (let dep of depsIndex.depsToFragments.keys()) {
+    precachedAssets.add(dep);
+  }
+  for (let depImports of depsIndex.fragmentToFullDeps.values()) {
```
justinfagnani: shouldn't only one of these loops be necessary? when is a dependency in fragmentToFullDeps not in depsToFragments?

src/service-worker.ts
```diff
@@ -0,0 +1,125 @@
...
+export interface AddServiceWorkerOptions {
+  project: PolymerProject;
+  buildRoot: string;
+  bundled?: boolean;
+  serviceWorkerPath?: string;
+  swConfig?: SWConfig;
+}
+
+/**
+ * Parses a depsIndex object and returns an array of file paths, pointing to
```
justinfagnani: "Parses" is the wrong word here Maybe "Takes a DepsIndex and returns an array of paths to precache"

src/service-worker.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Parses a depsIndex object and returns an array of file paths for a BUNDLED
+ * build, pointing to the project's required service worker precached assets.
+ */
+function getBundledPrecachedAssets(project: PolymerProject) {
+  let precachedAssets = new Set<string>(project.analyzer.allFragments);
+  precachedAssets.add(project.entrypoint);
+
+  return Array.from(precachedAssets).concat(project.bundler.sharedBundleUrl);
```
justinfagnani: might as well add the sharedBundleUrl to the precachedAssest like you do the entrypoint, so it's consistent

src/service-worker.ts
justinfagnani: `!!` means that you'd like the expression to pass for `true` and fail for `false`, and in this case there's really no such distinction. `!= null` means that it shouldn't be null or undefined, and while this lets non-string types though, it's (to me) a more accurate intent of the test. You could of course add a `typeof` on top of that.

test/test-project/gulpfile.js
```diff
@@ -48,5 +68,21 @@ gulp.task('test1', () => {
     // TODO(justinfagnani): allow filtering of files before writing
     .pipe(gulp.dest('build/unbundled'));
 
-  return mergeStream(bundled, unbundled);
+  return Promise.all([waitFor(bundled), waitFor(unbundled)]).then(() => {
```
justinfagnani: for the demo gulpfile, what do you think about using `.on('end')` instead of a Promise helper function?

test/test-project/gulpfile.js
```diff
@@ -48,5 +68,21 @@ gulp.task('test1', () => {
...
+    return Promise.all([
+      addServiceWorker({
```
justinfagnani: Let's submit as-is, but I think we can get to something like this:

```javascript
return Promise.all([
      addServiceWorker({
        projectConfig: project.unbundledConfig,
      }),
      addServiceWorker({
        projectConfig: project.bundledConfig,
      }),
    ]);
```

Where the project does the work of figuring out which assets comprise the project files, and what the output directory is.

justinfagnani: Actually, if we do the work of waiting for streams to finish in `addServiceWorker` we could do this:

```javascript
return Promise.all([
  addServiceWorker(project.unbundledConfig, unbundled),
  addServiceWorker(project.bundledConfig, bundled),
]);
```

src/service-worker.ts
FredKSchott: good point, this was taken from the old code so I assumed it was necessary. I'll test and fix or include a note if required for some reason.

src/service-worker.ts
FredKSchott: 👍 good call

test/test-project/gulpfile.js
FredKSchott: I'll use merge-stream so that we don't need to coordinate between two different event handlers.

test/test-project/gulpfile.js
FredKSchott: we'll just need to find a way to include the buildRoot in each config, and do it in a way that the user can still customize the call on top of this. Agreed that this should be treated as out of scope for now.

Lets keep that waiting logic outside of the function for now. I'm pushing a new change that waits for each individual buildPhase to create the service worker, instead of coupling them both together. Let me know what you think.

src/service-worker.ts
FredKSchott: yup, running:
`depImports.imports.forEach((s) => precachedAssets.add(s));` 

for each `fragmentToFullDeps` value creates the same Set as:
`depsIndex.depsToFragments.keys()`

src/service-worker.ts
FredKSchott: works for me!

FredKSchott:
@justinfagnani PTAL, specifically at the new `addServiceWorker()` coordination in the gulpfile.

test/test-project/gulpfile.js
```diff
@@ -33,20 +53,40 @@ gulp.task('test1', () => {
...
 
-  let unbundled = fork(allFiles)
+  let unbundledPhase = fork(allFiles)
     // write to the unbundled folder
     // TODO(justinfagnani): allow filtering of files before writing
     .pipe(gulp.dest('build/unbundled'));
 
-  return mergeStream(bundled, unbundled);
+  // Once the unbundled build stream is complete, create a service worker for the build
+  let unbundledPostProcessing = waitFor(unbundledPhase).then(() => {
```
justinfagnani: 👍

justinfagnani:
LGTM!

PR #9, opened 2016-07-27
no longer throws an error when no entrypoint or shell is given

FredKSchott:
/cc @justinfagnani

### DIFF
--- a/src/polymer-project.ts
+++ b/src/polymer-project.ts
@@ -112,8 +112,12 @@ export class PolymerProject {
 
   constructor(options?: ProjectOptions) {
     this.root = options.root || process.cwd();
-    this.entrypoint = osPath.resolve(this.root, options.entrypoint);
-    this.shell = osPath.resolve(this.root, options.shell);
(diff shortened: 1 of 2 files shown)


src/polymer-project.ts
```diff
@@ -112,8 +112,12 @@ export class PolymerProject {
 
   constructor(options?: ProjectOptions) {
     this.root = options.root || process.cwd();
-    this.entrypoint = osPath.resolve(this.root, options.entrypoint);
-    this.shell = osPath.resolve(this.root, options.shell);
+    if (options.entrypoint) {
+      this.entrypoint = osPath.resolve(this.root, options.entrypoint);
+    }
+    if (options.shell) {
+      this.shell = osPath.resolve(this.root, options.shell);
```
FredKSchott: do we have a preference as a team for using null for explicit "not set" properties?

web-padawan:
Thanks, this is related to the problem described in #8

justinfagnani:
LGTM

PR #12, opened 2016-07-27
add declaration property to tsconfig

FredKSchott:
`*.d.ts` files are now generated alongside the `*.js` files, in the lib directory.

/cc @justinfagnani am I missing anything? will this be enough for typescript to pick them up automatically?


justinfagnani:
You also need a `"typings"` entry in `package.json`

FredKSchott:
@justinfagnani done, PTAL

justinfagnani:
LGTM

PR #15, opened 2016-07-28
Remove unused files (old files from polymer-cli that are no longer needed)

FredKSchott:
This behavior now lives outside of polymer-build in the CLI. It can be pulled out of there in the future if desired.

/cc @justinfagnani


justinfagnani:
does `package.json` need to be updated?

FredKSchott:
Doh, dep-check to the rescue! No wonder tests were failing.

Just re-pushed, @justinfagnani ptal

justinfagnani:
LGTM :)

PR #16, opened 2016-07-29
Ignore remote files during build

### LINKED ISSUE #10: Error thrown during build when project has linked online resource
### Description
When I clone the current master branch and build it, then try to use it in my project, I get the following error, which is obviously caused by trying to read path on disk which is URL, so it fails

```
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: ENOENT: no such file or directory, open '/PATH/TO/MY/PROJECT/bower_components/font-roboto/https:/fonts.googleapis.com/css?family=Roboto:400,300,300italic,400italic,500,500italic,700,700italic'
    at Error (native)
```

### Versions & Environment
- polymer-build: 0.2.0 (current master branch)
- node: v6.3.1
- Operating System: Ubuntu 15.10

#### Steps to Reproduce
Try to build polymer project which uses 'font-roboto' bower package

#### Expected Results
No error thrown, links to online stuff should be ignored

#### Actual Results
Error is thrown as it tries to access online link on local disk

FredKSchott:
Fixes #10 

/cc @justinfagnani @filipproch

### DIFF
--- a/src/get-dependencies-from-document.ts
+++ b/src/get-dependencies-from-document.ts
@@ -8,14 +8,29 @@
  * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
  */
 
+import {parse as urlParse} from 'url';
 import {DocumentDescriptor} from 'hydrolysis';
 import {posix as posixPath} from 'path';
 import {Node, queryAll, predicates, getAttribute} from 'dom5';
+import * as logging from 'plylog';
+
+const logger = logging.getLogger('cli.build.get-dependencies');
 
 export interface DocumentDeps {
   imports?: Array<string>;
   scripts?: Array<string>;
   styles?: Array<string>;
+}
+
+/**
+ * Detects if a url is external by checking it's protocol. Also checks if it
+ * starts with '//', which can be an alias to the page's current protocol
+ * in the browser.
+ */
+function isDependencyExternal(url: string) {
+  // TODO(fks) 08-01-2016: Add additional check for files on current hostname
+  // but external to this application root. Ignore them.
+  return urlParse(url).protocol !== null || url.startsWith('//');
 }
 
 function collectScriptsAndStyles(tree: DocumentDescriptor): DocumentDeps {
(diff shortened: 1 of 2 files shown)


src/get-dependencies-from-document.ts
```diff
@@ -8,16 +8,29 @@
...
   styles?: Array<string>;
 }
 
+/**
+ * Detects if a url is remote by checking it's protocol. Also checks if it
+ * starts with '//', which can be an alias to the page's current protocol
+ * in the browser.
+ */
+function isDependencyRemote(url: string) {
+  return urlParse(url).protocol !== null || url.startsWith('//');
```
justinfagnani: shouldn't this also check the hostname?

justinfagnani: "remote" should include URLs on the same host that are outside the application root, but first we need to add support for application roots. Can you add a TODO here for that?

src/get-dependencies-from-document.ts
```diff
@@ -8,16 +8,29 @@
...
+const logger = logging.getLogger('cli.build.get-dependencies');
 
 export interface DocumentDeps {
   imports?: Array<string>;
   scripts?: Array<string>;
   styles?: Array<string>;
 }
 
+/**
+ * Detects if a url is remote by checking it's protocol. Also checks if it
```
justinfagnani: Let's call this "external" rather than "remote"

src/get-dependencies-from-document.ts
```diff
@@ -40,14 +53,37 @@ function collectScriptsAndStyles(tree: DocumentDescriptor): DocumentDeps {
...
   let allStyleDeps = new Set<string>();
-
   let deps: DocumentDeps = collectScriptsAndStyles(descriptor);
-  deps.scripts.forEach((s) => allScriptDeps.add(posixPath.join(dir, s)));
-  deps.styles.forEach((s) => allStyleDeps.add(posixPath.join(dir, s)));
+
+  // Collect all script dependencies
+  deps.scripts.forEach((scriptDep) => {
+    if (isDependencyRemote(scriptDep)) {
+      logger.debug('remote dependency ignored', {dep: scriptDep});
```
justinfagnani: not sure if logging this is that useful

justinfagnani:
LGTM with small comments

src/get-dependencies-from-document.ts
FredKSchott: Oh good call. Yea that's a much more complicated issue, I'll add the TODO.

src/get-dependencies-from-document.ts
FredKSchott: 👍

src/get-dependencies-from-document.ts
FredKSchott: Currently we're logging what happens to each dependency found via analysis. This has helped a few times with tracking down bad paths, but you're right we could be getting into overkill territory. Lets stick to the trend for now and then reevaluate if logs ever feel too noisy.

jpincas:
Is this working?

```
function isDependencyExternal(url: string) {
  // TODO(fks) 08-01-2016: Add additional check for files on current hostname
  // but external to this application root. Ignore them.
  return parseUrl(url).protocol !== null || url.startsWith('//');
}
```

Once of my imports is:

`<link rel="import" href="http://localhost:3000/admin-panel/polymer-imports.html">`

and it is causing a build error.  Doesn't seem to matter whether I start the href with the protocol or '//'.

jpincas:
OK - looking through the code a bit more closely I can see that ignoring external dependencies only applies to script and style dependencies.  **Is there a particular reason for this?** I am importing a dynamically rendered html file from an external server which contains a lot of setup information as behaviors.  It works perfectly with 'polymer serve' and seems like a good architecture.  Is it unwise for some reason?

PR #20, opened 2016-08-03
Handle external dependencies in hydrolosis resolver as well

FredKSchott:
Fixes https://github.com/Polymer/polymer-cli/issues/261 for real this time.

@garlicnation originally suggested a `NoopResolver` with string/regex matching, but I ultimately kept the logic in `StreamResolver` so that we could have more control over detecting what's external and whats not. 

/cc @justinfagnani


justinfagnani:
LGTM

PR #21, opened 2016-08-04
use absolute file paths in stream resolver to prevent duplicates

FredKSchott:
First step in solving some of our index.html absolute path woes. By internally tracking absolute file paths instead of url paths we get rid of a lot of potential ambiguity.

Also adds tests for a monolithic architecture app (no wonder our support has been so lacking :)

Fixes https://github.com/Polymer/polymer-cli/issues/337
/cc @justinfagnani

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -113,12 +113,11 @@ export class StreamAnalyzer extends Transform {
 
   _transform(file: File, encoding: string, callback: FileCB): void {
     let filePath = file.path;
-    let fileUrl = urlFromPath(this.root, file.path);
     this.addFile(file);
 
     // If our resolver is waiting for this file, resolve its deferred loader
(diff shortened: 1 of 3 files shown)


FredKSchott:
Holy 💩 this actually solves all of our index.html problems! Bundler is now running successfully with this change

src/analyzer.ts
```diff
@@ -117,8 +117,8 @@ export class StreamAnalyzer extends Transform {
     this.addFile(file);
 
     // If our resolver is waiting for this file, resolve its deferred loader
-    if (this.resolver.hasDeferredFile(fileUrl)) {
-      this.resolver.resolveDeferredFile(fileUrl, file);
+    if (this.resolver.hasDeferredFile(filePath)) {
```
justinfagnani: remove `fileUrl`?

justinfagnani:
LGTM!

PR #27, opened 2016-08-23
stop directly modifying the object passed to generateServiceWorker()

FredKSchott:
Fixes https://github.com/Polymer/polymer-cli/issues/345

This is much more serious than I thought, service-worker generation is currently broken in the CLI when a user defines their own `swConfig.staticFileGlobs` array.

/cc @justinfagnani Lets get this merged in ASAP along with the `polymer.json` CLI fix

### DIFF
--- a/src/service-worker.ts
+++ b/src/service-worker.ts
@@ -59,15 +59,17 @@ function getBundledPrecachedAssets(project: PolymerProject) {
  * contents), based off of the options provided.
  */
 export function generateServiceWorker(options: AddServiceWorkerOptions): Promise<Buffer> {
+  console.assert(!!options, '`project` & `buildRoot` options are required');
   console.assert(!!options.project, '`project` option is required');
   console.assert(!!options.buildRoot, '`buildRoot` option is required');
 
+  options = Object.assign({}, options);
   let project = options.project;
   let buildRoot = options.buildRoot;
-  let swConfig: SWConfig = options.swConfig || {};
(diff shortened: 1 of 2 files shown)


src/service-worker.ts
```diff
@@ -59,12 +59,14 @@ function getBundledPrecachedAssets(project: PolymerProject) {
...
   console.assert(!!options.buildRoot, '`buildRoot` option is required');
 
+  options = Object.assign({}, options);
   let project = options.project;
   let buildRoot = options.buildRoot;
-  let swConfig: SWConfig = options.swConfig || {};
+  let swConfig: SWConfig = Object.assign({}, options.swConfig);
 
   return project.analyzer.analyzeDependencies.then((depsIndex: DepsIndex) => {
     let staticFileGlobs = swConfig.staticFileGlobs || [];
```
justinfagnani: to make sure, you might want to do similar here:
```
let staticFileGlobs = Array.from(swConfig.staticFileGlobs || []);
```

justinfagnani:
LGTM

src/service-worker.ts
FredKSchott: we actually never modify `swConfig.staticFileGlobs`. Instead we return new arrays via concat and then overwrite the `staticFileGlobs` property with one of those new arrays. So this isn't an issue today.

BUT you're right, code changes and this might not be the case forever. I'll add this extra precaution back in.

PR #38, opened 2016-09-09
Mention whitelisting 3rd party services

robdodson:


robdodson:
k I think this is read. @fks PTAL

FredKSchott:
looks good, can you move down the section on 'sw-precache' to be immediately above this? Since they're related they'll be more helpful together. 

Either section could also be re-written to include a more explicit mention to "see [sw-preache](https://github.com/GoogleChrome/sw-precache#options-parameter) for a list of all supported options" (this should have probably been included from the beginning).

robdodson:
Updated PTAL

FredKSchott:
Awesome, LGTM!

PR #39, opened 2016-09-09
support relative root option

FredKSchott:
Previously, `root` only supported full system file paths. This made the option useless in a shared project. Now we support paths relative to the current working directory where the gulpfile is run.

/cc @robdodson @justinfagnani

### DIFF
--- a/src/polymer-project.ts
+++ b/src/polymer-project.ts
@@ -32,7 +32,9 @@ const extensionsForType: {[mimetype: string]: string} = {
 
 export interface ProjectOptions {
   /**
-   * Path to the root of the project on the filesystem.
+   * Path to the root of the project on the filesystem. This can be an absolute
+   * path, or a path relative to the current working directory. Defaults to the
+   * current working directory of the process.
    */
   root?: string;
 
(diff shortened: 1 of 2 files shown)


robdodson:
LGTM

chuckh:
LGTM2, you might want to update Readme to reflect this enhancement.

FredKSchott:
@chuckh good call, our readme points to the type definition in our source file, so I've updated the description there. I'll also add something to our official documentation as well once this is released.

src/polymer-project.ts
```diff
@@ -112,7 +114,11 @@ export class PolymerProject {
   bundler: Bundler;
 
   constructor(options?: ProjectOptions) {
-    this.root = options.root || process.cwd();
+    this.root = process.cwd();
```
justinfagnani: btw, this can be just: `this.root = osPath.resolve(options.root)` or perhaps `this.root = osPath.resolve(options.root || '')` to guard against null.

FredKSchott: TIL `path.resolve()` will resolve from the current working directory by default!

PR #43, opened 2016-09-15
Move polymer-build configuration logic to polymer-project-config

### LINKED ISSUE #36: excludeDependencies?
Can we include excludeDependencies option as part of the build? or is there a simpler way to exclude files from the build?

FredKSchott:
~~*Tests are expected to fail until polymer-project-config is published to npm*~~

This PR moves old configuration logic over to the new polymer-project-config library. See https://github.com/Polymer/polymer-project-config/pull/1 for background and bigger discussion on the move.

Fixes #36

/cc @justinfagnani


src/analyzer.ts
```diff
@@ -165,7 +145,7 @@ export class StreamAnalyzer extends Transform {
   }
 
   isFragment(file: File): boolean {
```
justinfagnani: seems like this should move into `ProjectConfig`

src/bundle.ts
```diff
@@ -101,30 +82,30 @@ export class Bundler extends Transform {
   }
 
   isEntrypoint(file: File): boolean {
```
justinfagnani: same here: move into `ProjectConfig`

src/bundle.ts
```diff
@@ -101,30 +82,30 @@ export class Bundler extends Transform {
...
         let newShellContent = this._addSharedImportsToShell(bundles);
         shellFile.contents = new Buffer(newShellContent);
       }
 
-      for (let fragment of this.allFragments) {
-        let fragmentUrl = urlFromPath(this.root, fragment);
-        let addedImports = (fragment === this.shell && this.shell)
+      for (let fragment of this.config.allFragments) {
+        let fragmentUrl = urlFromPath(this.config.root, fragment);
+        let addedImports = (fragment === this.config.shell && this.config.shell)
```
justinfagnani: for clarity let's move the `fragment === this.config.shell && this.config.shell` into a variable named `isShell`

src/bundle.ts
FredKSchott: 👍 great call

src/analyzer.ts
FredKSchott: 👍 great call

FredKSchott: this makes the "fragments" vs. "allFragments" collision even more confusing (https://github.com/Polymer/polymer-project-config/pull/1#r79076957)

justinfagnani: Yeah, super not happy with the naming... :(

FredKSchott:
@justinfagnani @rictic @usergenic PTAL

rictic:
Cool! I like the deduplication here. Thanks for ccing me in so I can get some more polymer-build context.

Leaving the review proper to @justinfagnani and @usergenic as they'll have better context right now though.

PR #44, opened 2016-09-16
Fix old naming of service-worker properties

### LINKED ISSUE #24: Minor nomenclature changes
I propose to rename these exposed variables:

- [`swConfig`](https://github.com/Polymer/polymer-build/blob/master/src/service-worker.ts#L25) to `serviceWorkerConfig` to maintain the same nomenclature as `serviceWorkerPath`.

- [`serviceWorkerPath`](https://github.com/Polymer/polymer-build/blob/master/src/service-worker.ts#L24) to `serviceWorkerFilename`. Because the real path is created at [src/service-worker.ts#L113](https://github.com/Polymer/polymer-build/blob/master/src/service-worker.ts#L113) with the `buildRoot` and the filename.

FredKSchott:
Fixes #24 

Nonbreaking, but does deprecate two `AddServiceWorkerOptions` properties:
- `serviceWorkerPath` -> `path`
- `swConfig` -> `swPrecacheConfig`

/cc @robdodson @abdonrd @justinfagnani


justinfagnani:
LGTM

PR #47, opened 2016-09-21
upgrade to typescript 2.0 and fix broken build process

FredKSchott:
@usergenic and I ran into problems with `gulp-typescript` yesterday, this upgrades to the latest 2.0 version and fixes some issues with config, custom_typings, etc.

/cc @usergenic @justinfagnani


justinfagnani:
yay! taking a look

test/service-worker_test.js
```diff
@@ -64,21 +64,30 @@ suite('service-worker', () => {
   suite('generateServiceWorker()', () => {
 
     test('should throw when options are not provided', () => {
-      assert.throws(() => {
-        serviceWorker.generateServiceWorker();
-      }, Error.AssertionError, '`project` & `buildRoot` options are required');
+      return serviceWorker.generateServiceWorker()
+        .then(() => { throw new Error('generateServiceWorker() resolved, expected rejection!'); })
```
justinfagnani: I would do Promise inversions as two callback to `then()`, so that you don't have to worry about the exception you throw in `then()` flowing into `catch()`. Then just assert.fail() in the first callback. Alternatively, we have a handy `invert()` function hangout out around here somewhere...

tsconfig.json
```diff
@@ -9,24 +9,16 @@
...
     "outDir": "lib",
     "rootDir": "src"
   },
-  "filesGlob": [
-    "node_modules/typescript/lib/lib.core.es7.d.ts",
-    "custom_typings/*.d.ts",
-    "src/**/*.ts",
-    "typings/index.d.ts"
-  ],
   "files": [
```
justinfagnani: TS 2.x supports `includes` and `excludes`. Use those instead of `files`.

typings.json
```diff
@@ -8,7 +8,6 @@
     "merge-stream": "registry:dt/merge-stream#1.0.0+20160313224417",
```
justinfagnani: can you migrate to `@types` while we're at this?

test/service-worker_test.js
justinfagnani: https://github.com/Polymer/polymer-analyzer/blob/master/src/test/test-utils.ts#L23

tsconfig.json
FredKSchott: 🎉 nice! It's behavior is a little weird, which is why I hadn't been able to get it working before.  Files still need to be referenced in the `files` array.

typings.json
FredKSchott: not all of them are supported yet, but I'll migrate the ones I can

usergenic:
TRAVIS Y U NO LIKE?

```
npm ERR! peerinvalid The package typescript@2.1.0-dev.20160921 does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer tslint@3.15.1 wants typescript@>=1.7.3
```

Anything we can do to fix that here too so we have green travis checks?

test/service-worker_test.js
FredKSchott: nice, I like this.

FredKSchott:
@usergenic the `typescript@next` dev versions were messing with semver matching, which was causing problems with npm2 installations. Switching back to the latest stable version (which we should be on, now that TS 2.0 is stable) fixes it.

FredKSchott:
@justinfagnani PTAL

tsconfig.json
justinfagnani: you shouldn't. We don't use `files` at all in the analyzer

typings.json
justinfagnani: which ones? All of these come from DT, so they should be available in `@types`

tsconfig.json
```diff
@@ -9,40 +9,18 @@
...
-    "src/path-transformers.ts",
+    "node_modules/typescript/lib/lib.es2016.d.ts",
     "typings/index.d.ts"
   ],
-  "atom": {
-    "rewriteTsconfig": true
-  }
+  "include": [
+    "custom_typings/**/*.ts",
+    "node_modules/@types/**/*.ts",
```
justinfagnani: I don't think you need or want this line. tsc automatically looks for these when you use node resolution

tsconfig.json
```diff
@@ -9,40 +9,18 @@
...
-    "src/service-worker.ts",
-    "src/path-transformers.ts",
+    "node_modules/typescript/lib/lib.es2016.d.ts",
     "typings/index.d.ts"
   ],
-  "atom": {
-    "rewriteTsconfig": true
-  }
+  "include": [
+    "custom_typings/**/*.ts",
```
justinfagnani: @rictic do you remember why we use `/// <reference path="../custom_typings/main.d.ts" />` in analyzer.ts rather than include `custom_typings` in tsconfig? Is it for declarations and dependers?

rictic: Yeah, it's for the generated `.d.ts` declarations file. If you only mention the dependency in your tsconfig then downstream users won't see the custom typings.

tsconfig.json
rictic: +1

gulpfile.js
```diff
@@ -81,7 +68,7 @@ gulp.task('eslint', () =>
     .pipe(eslint.failAfterError()));
 
 gulp.task('depcheck', () =>
-  depcheck(__dirname, {})
+  depcheck(__dirname, {ignoreMatches: ['@types/**']})
```
justinfagnani: can you comment the ignore?

test/service-worker_test.js
```diff
@@ -64,21 +64,36 @@ suite('service-worker', () => {
   suite('generateServiceWorker()', () => {
 
     test('should throw when options are not provided', () => {
-      assert.throws(() => {
-        serviceWorker.generateServiceWorker();
-      }, Error.AssertionError, '`project` & `buildRoot` options are required');
+      return serviceWorker.generateServiceWorker()
+        .then(
+          () => { throw new Error('generateServiceWorker() resolved, expected rejection!'); },
```
justinfagnani: instead of throwing, you can: `assert.fail('generateServiceWorker() resolved, expected rejection!')`

gulpfile.js
rictic: We should really make that part of the default ignoreMatches

test/service-worker_test.js
rictic: Any reason to prefer assert.fail over new Error?

justinfagnani: The UX. I think an uncaught error is reported different, and tries to display a stacktrace. Here we know it's a plain failure and the stack won't help.

gulpfile.js
justinfagnani: +1

tsconfig.json
FredKSchott: I'll add this for the `sw-precache` custom typing and we can look into publishing it directly to npm. All the others are internal only and shouldn't affect the `polymer-build` declarations

gulpfile.js
FredKSchott: +10, I'll create a PR and comment for now

test/service-worker_test.js
FredKSchott: I'm fine with either, the difference seems minor

FredKSchott:
@justinfagnani @rictic PTAL

gulpfile.js
FredKSchott: Already tracking here: https://github.com/depcheck/depcheck/issues/163

src/bundle.ts
```diff
@@ -104,68 +104,65 @@ export class Bundler extends Transform {
...
-      let promises: Promise<any>[] = [];
-
-      if (this.shell) {
-        let shellFile = this.analyzer.getFile(this.shell);
-        console.assert(shellFile != null);
-        let newShellContent = this._addSharedImportsToShell(bundles);
-        shellFile.contents = new Buffer(newShellContent);
-      }
+  async _buildBundles(): Promise<Map<string, string>> {
+    let bundles = await this._getBundles();
```
rictic: Here and elsewhere: preferred style is `const` by default, `let` only when necessary.

src/bundle.ts
```diff
@@ -104,68 +104,65 @@ export class Bundler extends Transform {
...
+    let sharedDepsBundle = (this.shell)
+        ? urlFromPath(this.root, this.shell)
+        : this.sharedBundleUrl;
+    let sharedDeps = bundles.get(sharedDepsBundle) || [];
+    let promises: Promise<any>[] = [];
```
rictic: Minimize use of `any`, it's usually a code smell. Would this work?

```typescript
const promises: Promise<{url: string, contents: string}>[] = []
```

tsconfig.json
```diff
@@ -4,45 +4,20 @@
     "module": "commonjs",
     "moduleResolution": "node",
     "isolatedModules": false,
-    "experimentalDecorators": true,
-    "emitDecoratorMetadata": true,
     "declaration": true,
     "noImplicitAny": true,
     "removeComments": false,
-    "noLib": true,
```
rictic: noLib still makes sense, or

`"lib": ["es2017"],`

typings.json
rictic: Only DT stuff in their 2.0 branch is in `@types`. They're using this as a chance to standardize on modern practices.

rictic:
As a code review style note, when possible instead of rewriting existing commits add --fixup commits instead, and you can rebase them right before going to master. It makes it easier for us to see only the new changes.

src/bundle.ts
FredKSchott: Agreed, but I'd like to keep the scope of this PR small for now. This diff was only a indentation change due to a move over to async-await above. I'll make a separate PR to address this.

src/bundle.ts
FredKSchott: Agreed, but I'd like to keep the scope of this PR small for now. I'll make a separate PR to address this across the repo once this has been merged.

FredKSchott:
@rictic TIL! I've never used the `--fixup` feature before, I'll play around with it.

@rictic @justinfagnani thanks for the help so far with all the different TypeScript config options, feeling much more confident in our config setup now.

test/service-worker_test.js
justinfagnani: I'd use `fail()`

tsconfig.json
justinfagnani: Please try to remove `files` before submitting. One file is replaced by compilerOptions.lib, the other should go in includes.

src/bundle.ts
justinfagnani: Agreed w/ the agree. In the future, just do the async/await in another PR also :)

tsconfig.json
FredKSchott: 👍 got it! Taking out`"typings/index.d.ts"` was what caused the error I was seeing, but it turns out it was a problem with typings and not typescript.

PR #48, opened 2016-09-22
move to polymer-analyzer

FredKSchott:
Continue to support hydrolysis due to (soon to be removed) vulcanize usage in `bundler.ts`.
Attempts to change as little existing behavior as possible. New improvements powered by the new analyzer will come in future PRs.

/cc @usergenic @justinfagnani @rictic


gulpfile.js
```diff
@@ -75,7 +75,10 @@ gulp.task('depcheck', () =>
       // "@types/*" dependencies are type declarations that are automatically
       // loaded by TypeScript during build. depcheck can't detect this
       // so we ignore them here.
-      ignoreMatches: ['@types/*']
+      // "hydrolosis" is being incorrectly matched because it is no longer
+      // used directly in code. TODO: remove from this list once we remove
```
justinfagnani: where's it being used?

src/analyzer.ts
```diff
@@ -20,12 +22,16 @@ import {Node, queryAll, predicates, getAttribute} from 'dom5';
 
 import {FileCB, VinylReaderTransform} from './streams';
 import {urlFromPath, pathFromUrl} from './path-transformers';
-import {DocumentDeps, getDependenciesFromDocument, isDependencyExternal}
-  from './get-dependencies-from-document';
 
 const minimatchAll = require('minimatch-all');
 const logger = logging.getLogger('cli.build.analyzer');
 
+export interface DocumentDeps {
```
justinfagnani: looks like maybe the only place that produces one of these includes all fields. Maybe make them non-optional.

src/analyzer.ts
```diff
@@ -111,40 +127,38 @@ export class StreamAnalyzer extends Transform {
     return this._dependenciesProcessingStream;
   }
 
-  _transform(file: File, encoding: string, callback: FileCB): void {
+  async _transform(file: File, encoding: string, callback: FileCB): Promise<void> {
```
justinfagnani: why is this `async`? `_transform` isn't intended to return anything. I'd consider using an AIIFE below to demark the async part of the work and use `await`.

src/analyzer.ts
```diff
@@ -186,11 +200,34 @@ export class StreamAnalyzer extends Transform {
...
+      if (isDependencyExternal(importUrl)) {
+        logger.debug(`ignoring external dependency: ${importUrl}`);
+      } else if (importDep.type === 'html-script') {
+        scripts.add(importUrl);
+      } else if (importDep.type === 'html-style') {
+        styles.add(importUrl);
+      } else if (importDep.type === 'html-import') {
+        imports.add(importUrl);
+      } else {
+        logger.warn(`unexpected import type encountered: ${importDep.type}`);
```
justinfagnani: I don't think a warn is necessary

src/analyzer.ts
```diff
@@ -186,11 +200,34 @@ export class StreamAnalyzer extends Transform {
...
+      }
+    }
 
-    return this.analyzer.metadataTree(url)
-        .then((tree) => getDependenciesFromDocument(tree, dir));
+    const deps = {
+      scripts: Array.from(scripts),
```
justinfagnani: any reason not to make these `Set`s?

src/analyzer.ts
```diff
@@ -238,11 +275,15 @@ export class StreamAnalyzer extends Transform {
...
+// TODO(fks) 09-21-2016: Remove once the move to polymer-analyzer is completed
+export interface BackwardsCompatibleUrlLoader
+  extends UrlLoader, HydrolysisResolver {};
+
+export class StreamLoader implements BackwardsCompatibleUrlLoader {
 
   root: string;
   analyzer: StreamAnalyzer;
-  deferredFiles = new Map<string, Deferred<string>>();
+  deferredFiles = new Map<string, (a: string) => string>();
```
justinfagnani: can you document this? What's the function do? Consider defining a type for it.

src/analyzer.ts
```diff
@@ -259,32 +300,50 @@ export class StreamResolver implements Resolver {
...
     let deferred = this.deferredFiles.get(filePath);
-    deferred.resolve(file.contents.toString());
+    deferred(file.contents.toString());
     this.deferredFiles.delete(filePath);
   }
 
-  accept(url: string, deferred: Deferred<string>): boolean {
-    logger.debug(`accept: ${url}`);
+  canLoad(url: string): boolean {
+    return true;
```
justinfagnani: Do you want to check for an external URL here?

src/analyzer.ts
```diff
@@ -259,32 +300,50 @@ export class StreamResolver implements Resolver {
...
+    return waitForFile;
+  }
+
+  /**
+   * Wraps the load() method to work in a way that is compliant with vulcanize
+   * & the old UrlResolver interface. To be removed once migration from
+   * hydrolosis to polymer-analyzer is complete.
+   */
+  accept(url: string, deferred: Deferred<string>): boolean {
+    this.load(url).then(deferred.resolve);
```
justinfagnani: maybe call `canLoad()` to ensure that the log is the same. (I know it's always true now)

gulpfile.js
FredKSchott: only type definitions for deprecations in `analyzer.ts`, which I think is why it can't find it. Looks like a bug in depcheck, I'll try to dig in a little when I get some time

src/analyzer.ts
FredKSchott: doh, good catch.

src/analyzer.ts
FredKSchott: This `DocumentDeps` is a deprecated type that we'll be able to get rid of in the new bundler logic. I kept the interface as-is for now so that we wouldn't have to refactor its current usage in `_addDependencies()`, `bundler.ts` & `service-worker.ts`. Is this worth adding a comment for to clarify?

src/analyzer.ts
FredKSchott: We want to match with external URLs so that we can resolve them as empty strings (otherwise we would error). In the future we'll probably want to refactor this into two loaders: 1 that handles external files and 1 that handles local. I'll add a TODO

FredKSchott:
@justinfagnani all comments addressed, PTAL

gulpfile.js
justinfagnani: ah, that makes sense. Don't worry about investigating, this is temporary.

src/analyzer.ts
rictic: +1 to commenting when we're doing things for backcompat reasons

PR #49, opened 2016-09-22
Convert "let" to "const" for default variable definitions

FredKSchott:
es6 christmas!

/cc @justinfagnani @rictic


rictic:
LGTM!

Did you intend to merge this without review?

FredKSchott:
I self-approved because it was so straight forward, we've always talked about doing it, and I really didn't want it to get stale. Definitely not getting into the habit :)

PR #51, opened 2016-09-23
add no-unused-variable rule to tslint

FredKSchott:
Now linting unused variables on lint. 

Since we have tslint we don't need to add this to tsconfig to warn on build as well, and I'd rather keep all of our linting rules in one place instead of two.

/cc @rictic @justinfagnani


tslint.json
```diff
@@ -10,6 +10,7 @@
         "no-internal-module": true,
         "no-trailing-whitespace": true,
         "no-var-keyword": true,
+        "no-unused-variable": true,
```
rictic: tsconfig > tslint

```
        "noUnusedLocals": true,
        "noUnusedParameters": true,
```

FredKSchott:
@rictic @justinfagnani PTAL

PR #54, opened 2016-09-26
add logic to print warnings during build analysis

FredKSchott:
- Using https://github.com/Polymer/polymer-analyzer/pull/338 for warning/error reporting.
- Collects and prints analysis warnings.
- Adds better error handling: now build will continue as long as some document is returned by the analyzer.
- Stops the build at the end of analysis if at least one `Severity.ERROR` warning was found
#### Example of new output:

screen shot 2016-09-30 at 2 28 12 pm

/cc @justinfagnani @rictic


test/analyzer_test.js
```diff
@@ -145,4 +145,46 @@ suite('Analyzer', () => {
...
+  // TODO(fks) 09-26-2016: The behavior of _getDependencies() will most likely
+  // change once vulcanize is converted to polymer-bunder. Hold off on implementing
+  // these tests until that work has been done.
+  suite.skip('_getDependencies()', () => {
+
+    test('will print warnings when analysis returns them', () => {
+      throw new Error('Not yet implemented');
+    });
+
+    test('will print a warning when analysis rejects with a warning-exception', () => {
```
rictic: s/warning-exception/WarningCarryingException/g ?

test/analyzer_test.js
```diff
@@ -145,4 +145,46 @@ suite('Analyzer', () => {
...
+    const root = path.resolve('test/static/project-analysis-error');
+    const analyzer = new StreamAnalyzer(
+      root,
+      path.resolve(root, 'index.html'),
+      null,
+      [],
+      [path.join(root, '**')]
+    );
+
+    mergeStream(
```
rictic: Consider faking stdout or stderr with a memory stream and adding a simple assert.matches to make sure that the error output includes the error you expect.

rictic: Consider faking stderr with a memory stream and adding a simple assert.matches to make sure that the error output includes the error you expect.

FredKSchott: Ended up removing WarningPrinter in favor of our own stripped down error reporting (due to line number issues, see #58).

FredKSchott:
@rictic @justinfagnani PTAL

rictic:
I'm sad that we lose track of the source of things and that this hurts our ability to report bugs. Have we looked into using source maps to keep track of this stuff?

As a shorter term option, are we doing any analysis over the code at the pre-transformed stage?

FredKSchott:
I agree, I brainstormed a few solutions in #58. I think running a first pass "linting" step would be ideal if the performance isn't too bad, because it would allow us to analyze both raw and processed source code. 

We could also consider changing the API to have the sources & dependencies split and be processed *after* analysis. This would be a pretty significant refactor, and I know @justinfagnani had some concerns when I pitched it back in the day, but now it could be the solution to this problem as well.

Regardless, there's no easy step that we could get done before summit, so this will have to do for now so that I can focus on other things.

src/analyzer.ts
```diff
@@ -52,6 +55,14 @@ function isDependencyExternal(url: string) {
   return parseUrl(url).protocol !== null || url.startsWith('//');
 }
 
+/**
+ * Get a longer, single-line error message for logging and exeption-handling
+ * analysis Warning objects.
```
rictic: It'd be good to document why we're not use WarningPrinter

rictic:
Makes sense. Leaving to @justinfagnani to approve.

rictic:
Oh, and I think you might need to use alpha.9 to get `Analyzer#analyze`

src/analyzer.ts
FredKSchott: great call, documentation ftw!

src/analyzer.ts
```diff
@@ -58,6 +58,11 @@ function isDependencyExternal(url: string) {
 /**
  * Get a longer, single-line error message for logging and exeption-handling
  * analysis Warning objects.
+ *
+ * Note: We cannot use WarningPrinter.printWarning() from the polymer-analyzer
+ * codebase because it's reported source ranges don't match the users source
```
rictic: nits (because I'm a crazy person): 
  `it's` -> `its`
  `users` -> `user's` or `users'`

FredKSchott:
@rictic @justinfagnani PTAL and re-read the OP. This includes some new features on top of the original PR.

justinfagnani:
I do think we should analyze for warnings on the original source, as I'm saying in #58. The difference between now and with Hydrolysis is that we have the ability in the future to handle alternate source in the Analzyer, where with Hydrolysis we had to make sure it ran post-compilation. There's not a ton of difference wrt source maps, since the Analyzer would just be hosting compilation and need source maps as well.

justinfagnani:
Can you rebase?

FredKSchott:
@justinfagnani I'm okay with moving analyzer to be in front of the optimizers in the CLI, as long as it all works. Can we mark that investigation as future work for after summit, so that it doesn't block this new error reporting?

src/analyzer.ts
justinfagnani: Add a TODO to test analysis before optimization. Once done we can use WarningPrinter because you can pass it a writable stream for output.

src/analyzer.ts
```diff
@@ -119,19 +135,33 @@ export class StreamAnalyzer extends Transform {
...
+          this.emit('error', error);
         }
       })();
     }
   }
 
   _flush(done: (error?: any) => void) {
+
+    this.printWarnings();
+    let allWarningCount = this.countWarningsByType();
```
justinfagnani: const

src/analyzer.ts
```diff
@@ -119,19 +135,33 @@ export class StreamAnalyzer extends Transform {
...
+    let errorWarningCount = allWarningCount.get(Severity.ERROR);
```
justinfagnani: const

src/analyzer.ts
```diff
@@ -172,11 +202,39 @@ export class StreamAnalyzer extends Transform {
...
+      } else if (warning.severity === Severity.WARNING) {
+        logger.warn(message);
+      } else {
+        logger.debug(message);
+      }
+    }
+  }
+
+  private countWarningsByType(): Map<Severity, number> {
+    let errorCountMap = new Map<Severity, number>();
```
justinfagnani: const

src/analyzer.ts
```diff
@@ -172,11 +202,39 @@ export class StreamAnalyzer extends Transform {
     this.files.set(urlFromPath(this.config.root, filepath), file);
   }
 
+  printWarnings(): void {
+    for (const warning of this.warnings) {
+      let message = getFullWarningMessage(warning);
```
justinfagnani: const

src/analyzer.ts
FredKSchott: Hmm, creating a new stream that uses our logger just for analysis seems a little round-a-bout, passing a function would be an easier interface to use as a consumer.

Will create a seperate issue to discuss.

PR #71, opened 2016-11-01
Update out of date dependencies

FredKSchott:
Updated:
- polymer-analyzer
- multipipe
- dom5
- fs-extra

Added:
- parse5


package.json
```diff
@@ -21,26 +21,28 @@
...
-    "fs-extra": "^0.30.0",
+    "dom5": "^2.0.1",
     "hydrolysis": "^1.23.3",
     "minimatch-all": "^1.0.2",
-    "multipipe": "^0.3.0",
+    "multipipe": "^1.0.2",
+    "parse5": "^2.2.2",
     "plylog": "^0.4.0",
-    "polymer-analyzer": "2.0.0-alpha.10",
+    "polymer-analyzer": "^2.0.0-alpha.17",
```
rictic: consider `=2.0.0-alpha.17` as we do breaking changes in alpha releases

FredKSchott: Ah good call. FWIW I don't think `^` is doing anything in npm when you have special `-alpha` tags like this one.

rictic:
I'm not 100% on it, but it did seem like ^ was getting later alpha releases of the analyzer in another repo I was working on.

FredKSchott:
Ah I could be wrong then. I thought I'd seen the opposite :)

PR #73, opened 2016-11-03
export missing typedefs

FredKSchott:
This way polymer-cli & other upstream TS projects can get rid of their `custom_typings` definition.

/cc @justinfagnani @rictic


FredKSchott:
Great tip!

So I've never been a fan of asking users to reach into your package file system to get what they need. It can create a lot of confusion about what's private (to the package) and what's public, which makes cutting new non-breaking versions a nightmare because basically everything becomes public once you have enough users. It also makes the "preferred" API harder to find vs. a whitelisted interface.

But on the other hand, I totally understand why that open access makes a lot of sense in a large project like the analyzer, since there's so many different kinds of data being passed around that exposing it all top-level would be messy and hard to maintain. Plus, the package file tree is very well organized so it's less messy for the consumer to navigate.

This might be worth talking about as a team as we get closer to v1.0 releases. 

A best-of-both-worlds compromise might be to export 1 public API for usage, but to encourage navigating the package file system to grab the type definitions you need.

PR #74, opened 2016-11-04
Add .vscode/settings & clang-format

FredKSchott:
No open PRs, so we can add this and format it all in the same PR.

/cc @rictic @justinfagnani


FredKSchott:
@rictic uh oh, looks like it's unhappy with our clang-formatting...

Do we even need eslint now that we have clang-format? I vote to remove, wdyt?

rictic:
Unsure on the rules we have set, but eslint can check things clang-format won't fix.

Almost always you can appease both, looks like we just need to add some commas?

FredKSchott:
@rictic sounds good, I kept eslint and ignored `comma-dangle` issues now that clang-format is handling when to add/remove those automatically. PTAL

rictic:
AFAIK, clang-format doesn't insert or remove commas. I think the correct fix will be to add those commas manually.

FredKSchott:
doh, I assumed because `npm run format` removed them, clang would remove them always. Re-submitted with comma-dangle enforced again and eslint passing

rictic:
On mobile so I can't track the diff effectively, but is it possible that npm run format didn't remove them, it just changed the whitespace so that they were required?

FredKSchott:
I see

gulpfile.js
```diff
@@ -37,44 +40,43 @@ gulp.task('build', ['clean'], () => {
...
-      // "hydrolosis" is being incorrectly matched because it is no longer
-      // used directly in code. TODO: remove from this list once we remove
-      // as dependency.
-      ignoreMatches: ['@types/*', 'hydrolysis']
-    })
-    .then((result) => {
+gulp.task(
+    'depcheck',
+    () => depcheck(__dirname, {
+            // "@types/*" dependencies are type declarations that are
```
justinfagnani: what format rule is causing this to be indented 8 spaces?

justinfagnani: I think we might want to stop using arrow function expressions for gulp task bodies. If we give them bodies and a return statement, clang might like it more.

FredKSchott: It seems like clang-format doesn't understand arrow functions (and might be treating them like objects?). I've tried playing around with configuration but the closest I've gotten is turning the indentation on/off or changing around the indentation widths. When converted to a true function, it fixes to this:

```js

gulp.task('tslint', function() {
  gulp.src('src/**/*.ts')
      .pipe(tslint({configuration: 'tslint.json', formatter: 'verbose'}))
      .pipe(tslint.report())
});
```

@rictic any ideas here?

rictic: I suspect that the misformatting is due to a complex function body in an arrow function without curly braces. I also find that style hard to parse correctly as a human.

Does it format this style better?

```javascript
gulp.task('depcheck', () => { 
  depcheck(__dirname, {
    // ...
  });
});
```

FredKSchott: Doesn't look like it: 

```js
gulp.task(
    'depcheck',
    () => {return depcheck(__dirname, {
                    ignoreMatches: ['@types/*', 'hydrolysis']
                  })
               .then((result) => {
                  // code goes here
               })});
```

FredKSchott: converted to normal functions for now

PR #75, opened 2016-11-08
remove redundant pushDependency() call after analysis

FredKSchott:
Found this as a part of new analyzer architecture, this change will simplify the architecture and remove a ton of debug noise.

Dependencies are loaded during analysis to create the entire dependency tree. They are loaded by calling `pushDependency()` inside the Analyzer.loader. But after this, we've been calling `pushDependency()` again on each dependency. This may have been necessary in the old analyzer, but it is no longer necessary in the new one.


rictic:
I see. Was anything downstream of this doing deduping? Was this causing some files to be processed many times?

FredKSchott:
`pushDependency` was doing the deduping internally, and it would exit and log a debug message when it did. Because every dependency had already been pushed during loading, this was logging a debug message and exiting for every single dependency.

Rolled in to #77

PR #77, opened 2016-11-10
Run analyzer first

FredKSchott:
- [x] README.md has been updated

## Background

The polymer-build libraries relies on the polymer-analyzer to function. Analysis was designed to run inside the build stream, after any optimization and pre-processing was run on the application source files. This was meant to support any optimizations that moved, combined, or created new files in the build. 

However we are now running into issues where minification and optimizations create un-analyzable code which results in errors during build. For example, JavaScript minification can strips important tags from code that are needed for detecting Polymer behaviors.   

## Objective

This PR moves analysis to the front of the build stream, which results in a simplified user interface for the library and more accurate code analysis. The analyzer no longer needs to be included by the user during build, while still filling the `sources` & `dependencies` streams automatically.

See the new additions to the [CHANGELOG](CHANGELOG.md) & [README](README.md) for more info.

/cc @rictic @justinfagnani @usergenic


README.md
```diff
@@ -60,20 +60,20 @@ Returns a readable stream of your project's source files. By default, these are
 
 Returns a readable stream of your project's dependencies. This stream is automatically populated based on the files loaded inside of your project. You can include additional dependencies via the `extraDependencies` property in [`ProjectOptions`](src/polymer-project.ts) (this can be useful when the analyzer fails to detect a necessary dependency.)
 
-#### project.analyzer
+#### project.startBuild()
```
FredKSchott: There's a description of what's happening in the analyzer: https://github.com/Polymer/polymer-build/blob/fbe949bbe39ecc5de40ff8edf036879752bc94fd/src/analyzer.ts#L119

But it doesn't go out of its way to answer this question. I'll reword to be more clear and then reference that in `project.startBuild()`.

src/analyzer.ts
```diff
@@ -95,36 +98,65 @@ export class StreamAnalyzer extends Transform {
...
+   * objects for each source.
    */
-  get dependencies(): Transform {
-    return this._dependenciesProcessingStream;
+  get sources(): NodeJS.ReadableStream {
+    return this._sourcesOutputStream;
   }
 
-  _transform(file: File, _encoding: string, callback: FileCB): void {
+  async analyzeFile(file: File): Promise<void> {
```
FredKSchott: nope, I'll make both this and _done private

FredKSchott:
@justinfagnani Refactored to remove the need for `start()`/`startBuild()` methods. PTAL

FredKSchott:
rebased to fix conflicts

src/analyzer.ts
```diff
@@ -72,14 +73,78 @@ function getFullWarningMessage(warning: Warning): string {
...
+    } catch (err) {
+      callback(err);
+      return;
+    }
+    callback(null, file);
+  }
+}
+
+/**
+ * A stream to analyze ever file that passes through it. This is used to analyze
```
justinfagnani: "ever" -> "every"

src/analyzer.ts
```diff
@@ -95,90 +160,125 @@ export class StreamAnalyzer extends Transform {
...
+  /**
+   * Return _sourcesOutputStream, which will contain fully loaded file
+   * objects for each source after analysis.
+   */
+  get sources(): NodeJS.ReadableStream {
+    return this._sourcesProcessingStream;
+  }
+
+  /**
+   * Resolve a loaded file for our analyzer's loader so that analysis is able to
```
justinfagnani: maybe "so that analysis is able to continue" -> "so that the analyzer can receive a requested file"?

src/analyzer.ts
```diff
@@ -72,14 +73,78 @@ function getFullWarningMessage(warning: Warning): string {
   }`;
 }
 
-export class StreamAnalyzer extends Transform {
-  config: ProjectConfig;
+/**
+ * A stream that tells the BuildAnalyzer to resolve each file it sees. It's
+ * important that resolving loaded files happens as soon as possible, in a way
```
FredKSchott: I tried to imply this with the second half of the sentence, but will make it more explicit that blocking analysis and other tasks is very, very bad :)

src/analyzer.ts
```diff
@@ -72,14 +73,78 @@ function getFullWarningMessage(warning: Warning): string {
...
+ * that doesn't block analysis or other tasks that rely on file loading.
+ */
+class ResolveStream extends Transform {
```
FredKSchott: sure, I like it. Changing `AnalyzeStream`->`AnalyzeTransform` as well

FredKSchott:
Thanks @justinfagnani!

PR #87, opened 2016-12-09
update merge-stream to v1.0.1, properly propagate errors

FredKSchott:


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 * **Interface Update!** `PolymerProject.analyzer` is no longer a required step in your build pipeline. Instead, analysis happens automatically while it fills the `project.sources()` and `project.dependencies()` streams with your project files. See [the README](/README.md) for updated examples of what build streams look like without the analyzer.
 * `StreamAnalyzer` is now `BuildAnalyzer` (since it is no longer a stream). This change only affects uses who are importing and/or using `StreamAnalyzer` directly from the `polymer-build` module.
+* **[`merge-stream`](https://github.com/grncdr/merge-stream) users:** Update to v1.0.1 or later if you are using `merge-stream` with `poylmer-build`. Stream errors do not propagate properly in previous versions of the library, and your build task may silently fail as a result.
```
rictic: typo: poylmer-build -> polymer-build

rictic: Is it the case that an unusually large number of our users are using merge-stream? It's slightly odd that we're rebroadcasting a change in a largely unrelated library

FredKSchott: Ah , good catch on the typo.

FredKSchott: Yea, it's the tool we recommend using to merge the two `sources()` and `dependencies()` streams together after optimization. It's mentioned in several places in the README. 

This is only relevant because of the analyze-first interface change. Since the analyzer was removed as an explicit step, we now propagate analysis errors directly into those two streams. This exposed a bug in <v1.0.1 of merge-stream, where errors weren't propagated through the merge-stream.

rictic: Dumb question, but what does it mean to propagate an error down a stream like this? Is this like a node event emitter `error` event? Or are we sending down `Warning` objects, or string representations of them or…

FredKSchott: In polymer-build, we print all warnings and then emit a single error if any of them were of type "ERROR" (https://github.com/Polymer/polymer-build/blob/536c7a846d1f854a7f244695c001af0bc853573a/src/analyzer.ts#L258)

PR #92, opened 2016-12-20
Convert tests to typescript

FredKSchott:
I have a big testing improvement ready to go for the bundler, but it uses TS. So this PR converts all of our tests to TS, and then that PR will follow with the testing improvements.


src/test/analyzer_test.ts
```diff
@@ -12,15 +12,17 @@
...
-const assert = require('chai').assert;
-const path = require('path');
-const BuildAnalyzer = require('../lib/analyzer').BuildAnalyzer;
-const waitForAll = require('../lib/streams').waitForAll;
-const sinon = require('sinon');
-const ProjectConfig = require('polymer-project-config').ProjectConfig;
-const Writable = require('stream').Writable;
+import {assert} from 'chai';
+import * as path from 'path';
+import File = require('vinyl');
```
usergenic: `npm install "@types/vinyl" --save-dev`?

src/test/analyzer_test.ts
```diff
@@ -42,7 +44,7 @@ suite('Analyzer', () => {
 
     test('fragment to deps list has only uniques', () => {
       const config = new ProjectConfig({
-        root: `test/static/analyzer-data`,
+        root: `src/test/static/analyzer-data`,
```
usergenic: This feels weird to me now because the static test assets are copied into `lib` but the test is loading them from `src`.

Fix my cognitive dissonance please!

src/test/analyzer_test.ts
```diff
@@ -231,14 +233,14 @@ suite('Analyzer', () => {
     const analyzer = new BuildAnalyzer(config);
 
     // Check that data isn't flowing through sources until consumer usage
-    assert.isNull(analyzer.sources._readableState.flowing);
+    assert.isNull((<any>analyzer.sources)._readableState.flowing);
```
usergenic: Why `<any>`?

usergenic:
Can you note why you decided to move the static `test` assets into `src/test`?

They're copied into `./lib` but then in a weird twist they also don't work without `./src` because paths in the test code.  Perhaps we can not copy them into `./lib` at all or maybe we can leave them in `./test`?

What do you think?

src/test/analyzer_test.ts
FredKSchott: Good catch, I should add a comment here. 

So the official Node TS definition doesn't define the `_readableState` property on streams because it is private. However, the official Node.js documentation documents this property [extensively](https://nodejs.org/api/stream.html#stream_three_states), and specifically as a way to check whether a stream is flowing or not. So we're stuck with TS telling us the property doesn't exist, and Node.js docs telling us it does and should be used in this situation.

We could also solve this by extending the Node typedef in `custom_typings/`. But since we only need it in one or two tests, scoping the change here with `<any>` seemed more appropriate.

src/test/analyzer_test.ts
FredKSchott: We already do :/ But because `vinyl` exports a class and not a true module (at least according to it's typedef) we can't use the `import ... from ...` syntax and need to continue using `require()`.

This is common-ish so I don't think it needs a comment, just one of those weird parts of TS + ES syntax.

FredKSchott:
@usergenic That's a good point about `test/static`. I had just naively moved the entire directory into `src/` but leaving those static files at project level would put them in one place (not `src/` & `lib/`) and simplify the build step (it needed a special step to copy non-ts files). Let me see if that works instead.

FredKSchott:
👍 Yup moving them back to top level definitely simplifies things. Let me see if that's an improvement we want to make to polymer-analyzer as well.

Thanks for the review @usergenic, PTAL

src/test/analyzer_test.ts
```diff
@@ -12,15 +12,17 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-'use strict';
+/// <reference path="../../node_modules/@types/mocha/index.d.ts" />
```
justinfagnani: I haven't had to do this in other .ts tests, are you sure it's necessary?

FredKSchott: I know I was surprised too. Only one reference is needed since it's treated as a global, but since there's no one entry-point to our tests suite adding it to all seemed safest in the event that files are moved around and/or refactored.

(ex: https://github.com/Polymer/polymer-analyzer/blob/master/src/test/analyzer_test.ts)

justinfagnani: Weird, because in other projects I don't have a reference to it at all. What tsc are you on?

FredKSchott:
```
polymer-build@0.6.0-alpha.3 /Users/fschott/Code/polymer-build
└── typescript@2.1.4 
```

Hmm, after some googling it looks like you can pass an option to your tsconfig to reference mocha automatically: http://stackoverflow.com/a/40197672 This means TypeScript will also mistakenly think `suite` and `test` are available in your library code as well, but I don't think that's a problem we'd ever encounter normally. 

I'm fine with adding this to our `tsconfig` and removing the explicit reference, if there aren't any objections? We could do the same in polymer-analyzer as well.

justinfagnani: I'm honestly not sure it's better to take it out, I was more curious why some projects didn't need it. Considering I'd rather just import it, it's probably better to leave it. Importing doesn't work, does it?

FredKSchott: Correct. mocha doesn't export suite/test, it adds them to the global namespace.

rictic:
Looks like there may be a race in calling `gulp test lint` between compiling and running depcheck: https://travis-ci.org/Polymer/polymer-build/jobs/186184230

FredKSchott:
@rictic arg you're right 😞 that's a quick fix, will make a separate PR

FredKSchott:
all comments addressed, PTAL (@usergenic especially since you still have changes requested)

src/test/util.ts
```diff
@@ -0,0 +1,22 @@
...
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+export function getFlowingState(stream: NodeJS.ReadableStream): boolean {
+  // Cast our streams to <any> so that we can check the flowing state.
+  // _readableState is undocumented in the Node.js TypeScript definition,
+  // however it is the supported way to assert if a stream is flowing or not.
+  // See: https://nodejs.org/api/stream.html#stream_three_states
+  const privateReadableState = (<any>stream)._readableState;
+  return privateReadableState.flowing;
```
justinfagnani: Nice to factor this out, but I'd go even a step further and make it more of an assertion in itself:
 * Rename to `isFlowing`
 * Use it in a plain assert: `assert(isFlowing(stream))`
 * Since`privateReadableState.flowing` is null sometimes: `return privateReadableState.flowing == true`

This is short of making an actual Chai plugin, but that's a lot more work

test-fixtures/bundle-project/index.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
```
justinfagnani: what's the reason for the new fixtures? Are they logically part of the TS conversion?

tsconfig.json
```diff
@@ -17,6 +19,6 @@
   },
   "include": [
     "custom_typings/**/*.ts",
-    "src/**/*.ts"
+    "src/**/*"
```
justinfagnani: revert this

test-fixtures/bundle-project/index.html
FredKSchott: Good catch, these were for the tests that I'd gone in to fix originally. Looks like the fixtures got sucked into this PR without the tests. Pulling these out...

src/test/util.ts
FredKSchott: Nice idea, but flowing state can actually be 3 values (`true`, `false`, `null`). see: https://nodejs.org/api/stream.html#stream_three_states

While we're currently only testing "hasn't started yet" and "has started and is flowing", we'll want to be able to test "has started and is stopped" (`false`) as well. I could still pull the assert into here, but given the 3 states I prefer writing out the assertion in the test itself.

FredKSchott:
@justinfagnani PTAL

PR #98, opened 2016-12-22
fix race condition in prepublish script

FredKSchott:
Fixes flakey installs and travis builds

/cc @rictic


rictic:
Yep, AFAIK this is the only way to fix this.

Use gulp they said. Run your tasks in parallel they said.

Maybe I'll write a task runner over the holidays…

FredKSchott:
Gulp 4 is supposed to fix this. 
Gulp 4 has been in development for the last 35 years.

Issue #97, opened 2016-12-21
Since 0.6.0-alpha.1 analyzing does not run in determined order


web-padawan:
### Description
After implementing "analyzer first" approach, unable to use  `polymer-build` as a step in gulp pipeline.
Looks like analysis ignores gulp tasks order at all.

### Versions & Environment
- polymer-build: 0.6.0-alpha.1, 0.6.0-alpha.3
- node: 6.9.1
- Operating System: Mac OS X El Capitan, Windows 7

#### Steps to Reproduce

See [example repo](https://github.com/web-padawan/polymer-sandbox): 

```
git clone https://github.com/web-padawan/polymer-sandbox
npm i
bower i
gulp build
```

For working example, please check [polymer-build-0.5.1](https://github.com/web-padawan/polymer-sandbox/tree/polymer-build-0.5.1) branch.

#### Expected Results
Bundling starts after beginning corresponfing step and ends.

#### Actual Results
Bundling starts before `vulcanize` step beginning and does not finish.
Same can be reproduced with `gulp 4` as well.

@FredKSchott am I doing something wrong here?

FredKSchott:
Thanks for reporting. It looks like polymer-build is just pre-loading your source files (that's the output you're seeing with verbose logging on). Nothing is being analyzed or or bundled until the bundling step is started, as expected.

What kind of problems is this causing for you?

web-padawan:
Actually bundling looks completed with `0.6.0-alpha.3`, but in fact shell does not include some dependencies, e. g. `polymer.html`.

Please compare: 0.6.0-alpha.3 output
```
[21:56:23] Finished 'vulcanize' after 1.82 s
[21:56:23] Starting 'minify:html'...
[21:56:23] minify:html all files 3.2 kB
[21:56:23] Finished 'minify:html' after 32 ms
[21:56:23] Starting 'crisper'...
[21:56:23] crisper all files 3.35 kB
[21:56:23] Finished 'crisper' after 21 ms
[21:56:23] Starting 'minify:js'...
[21:56:23] minify:js all files 395 B
[21:56:23] Finished 'minify:js' after 33 ms
[21:56:23] Starting 'finalize'...
[21:56:23] finalize all files 3.01 kB
[21:56:23] Finished 'finalize' after 13 ms
[21:56:23] Finished 'build' after 3.03 s
```

0.5.1 output
```
[22:01:09] Finished 'vulcanize' after 3.83 s
[...]

PR #99, opened 2016-12-22
don't load files until sources() or dependencies() is called

FredKSchott:
Brought up in: https://github.com/Polymer/polymer-build/issues/97#issuecomment-268619992

Currently, files begin loading from disk the second the PolymerProject instance is created. This is problematic for a build library because the files you need to load may not exist at the start of the build. They could be moved or even generated as a part of the build process, before polymer-build is started. 

It's the setup of a listener on the vinyl source stream (`_sourcesStream`) that starts the actual loading. So this PR moves all stream setup from the constructor (at `PolymerProject` creation-time)  to the time of their first use. 

Fixing this was an original goal for https://github.com/Polymer/polymer-build/pull/77, but it was dropped to keep that PR scope small. 


src/test/analyzer_test.ts
```diff
@@ -232,16 +233,63 @@ suite('Analyzer', () => {
     });
 
     const analyzer = new BuildAnalyzer(config);
+    assert.isFalse(analyzer.started);
+    analyzer.sources().pipe(new NoopStream());
+    assert.isTrue(analyzer.started);
+  });
+
+
+  test('calling dependencies() starts analysis', () => {
```
FredKSchott: These three tests (two new and 1 updated) are the only testing changes made by me. The rest of this `test/` diff is due to clang-formatting after the `.sources -> .sources()` & `dependencies -> dependencies()` change.

justinfagnani:
Isn't there a way for a ReadableStream implementor to be notified when it has 'data' listeners? We don't want to start producing data when someone _requests_ one of the file streams, but when they've hooked it up to a destination.

src/analyzer.ts
```diff
@@ -198,15 +214,17 @@ export class BuildAnalyzer {
    * Return _dependenciesOutputStream, which will contain fully loaded file
    * objects for each dependency after analysis.
    */
-  get dependencies(): NodeJS.ReadableStream {
+  dependencies(): NodeJS.ReadableStream {
```
FredKSchott: startAnalysis() begins the loading of files from disk. This was a significant enough side effect to warrant the change. And as a little bonus, it also now matches the `polymerProject.dependencies()` method that most users interact with.

FredKSchott:
@justinfagnani not that I've been able to find. The closest is the `pipe` event, but that is emitted on the destination stream and not the source stream. While I would prefer that behavior as well, the only way that I can see is by extending the Stream class to add our own `pipe()` logic, or by making the user call `startAnalysis()`/`start()`  explicitly to start the stream. 

But the only way to run into the issue you're describing is to call `polymerProject.sources()`/`polymerProject.dependencies()` and then not `pipe()` or connect the stream on that same tick. I've never seen any of our users do this before, and I can't think of a scenario where you'd ever need to. Given that, I don't think this is worth prioritizing right now (or at the very least, I don't think it is worth blocking this improvement over the current logic).

FredKSchott:
@justinfagnani @rictic PTAL

justinfagnani:
I'm not suggesting blocking this, just wondering if it's possible given that that's the behavior we want.

src/analyzer.ts
justinfagnani: I agree with that reasoning there, even though the breaking change and churn is unfortunate.

justinfagnani: Oh, I see this is on analyzer anyway, which isn't often the public interface. I take back the churn comment.

PR #101, opened 2017-01-04
[RELEASE] v0.6.0

FredKSchott:

The latest changes have been baking for a few weeks now, and all feedback so far has been positive. There's no upcoming work left that would block the big improvements contained in this version.

/cc @justinfagnani @rictic

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,20 +7,33 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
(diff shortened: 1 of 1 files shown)


FredKSchott:
I tried something new to display both the major and pre-release changes in the changelog, but the effect is kind of hidden in the diff. Check out the new changelog to see what it looks like and let me know what you think: https://github.com/Polymer/polymer-build/blob/prepare-release/v0.6.0/CHANGELOG.md

justinfagnani:
I like that change in the changelog!

PR #103, opened 2017-01-08
Fix CSS split/rejoin

tony19:
This addresses the review comments from #40, including:
 * Change `<style src="...">` (invalid) to `<link rel="stylesheet" href="...">`
 * Add tests

/cc @hurst @FredKSchott @justinfagnani

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 * `BuildAnalyzer.sources` & `BuildAnalyzer.dependencies` are now `BuildAnalyzer.sources()` & `BuildAnalyzer.dependencies()`, respectively. This change only affects uses who are importing and/or using the `BuildAnalyzer` class directly.
 * Fix issue where files were being loaded immediately, before the build stream was started.
+* Fix CSS split/rejoin (#32)
 
 
 ## [0.6.0] - 2017-01-14
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,16 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
-<!-- List New Changes Here -->
-
-## [0.7.0] - 2017-01-12
-
 * `BuildAnalyzer.sources` & `BuildAnalyzer.dependencies` are now `BuildAnalyzer.sources()` & `BuildAnalyzer.dependencies()`, respectively. This change only affects uses who are importing and/or using the `BuildAnalyzer` class directly.
 * Fix issue where files were being loaded immediately, before the build stream was started.
-* Fix CSS split/rejoin (#32)
 
 
-## [0.6.0] - 2017-01-04
+## [0.6.0] - 2017-01-14
 
 * **Interface Update!** `PolymerProject.analyzer` is no longer a required step in your build pipeline. Instead, analysis happens automatically while it fills the `project.sources()` and `project.dependencies()` streams with your project files. See [the README](/README.md) for updated examples of what build streams look like without the analyzer.
 * **[`merge-stream`](https://www.npmjs.com/package/merge-stream) users:** Update to v1.0.1 or later if you are using `merge-stream` with `polymer-build`. Stream errors do not propagate properly in previous versions of the library, and your build task may silently fail as a result.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+<!-- List New Changes Here -->
+
+## [0.7.0] - 2017-01-12
+
 * `BuildAnalyzer.sources` & `BuildAnalyzer.dependencies` are now `BuildAnalyzer.sources()` & `BuildAnalyzer.dependencies()`, respectively. This change only affects uses who are importing and/or using the `BuildAnalyzer` class directly.
 * Fix issue where files were being loaded immediately, before the build stream was started.
+* Fix CSS split/rejoin (#32)
 
 
-## [0.6.0] - 2017-01-14
+## [0.6.0] - 2017-01-04
 
 * **Interface Update!** `PolymerProject.analyzer` is no longer a required step in your build pipeline. Instead, analysis happens automatically while it fills the `project.sources()` and `project.dependencies()` streams with your project files. See [the README](/README.md) for updated examples of what build streams look like without the analyzer.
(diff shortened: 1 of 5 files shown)


tony19:
The test failures are due to the dom5 bug (which should be fixed by https://github.com/PolymerLabs/dom5/pull/46).

FredKSchott (commented):
@tony19 awesome! Thanks for working on this, looks really good.

Lets get that dom5 bug fixed and published and then I'll take a deeper look at the code.

Also adding @justinfagnani since he wrote a lot of this.

tony19:
@FredKSchott Rebased and brought in the dom5 bugfix for integration tests.

justinfagnani:
@tony19 I just published 2.1 of dom5 with the template traversal

tony19:
@justinfagnani Thanks, updated.

tony19:
There's a bug somewhere that prevents CSS rejoining for `build/unbundled`. I'll need to investigate.

CHANGELOG.md
```diff
@@ -7,11 +7,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+<!-- List New Changes Here -->
+
+## [0.7.0] - 2017-01-12
```
FredKSchott: Keep the changelog versions as is on master, we'll make sure it gets merged correctly when we cut the next release.

src/polymer-project.ts
```diff
@@ -244,6 +222,69 @@ class HtmlSplitter extends Transform {
       callback(null, file);
     }
   }
+
+  _splitScript(file: File, doc: parse5.ASTNode, filePath: string) {
+    const scriptTags = dom5.queryAll(doc, HtmlSplitter.isInlineScript);
+    for (let i = 0; i < scriptTags.length; i++) {
```
FredKSchott: can you clean up any differences between `splitScript()` & `splitStyle()`? For example, `splitScript` is using a `for()` loop while `splitStyle()` is using `styleTags.forEach()`.

src/test/polymer-project_test.ts
```diff
@@ -147,45 +147,166 @@ suite('PolymerProject', () => {
...
+    suiteSetup((done) => {
+      defaultProject.sources()
         .pipe(defaultProject.splitHtml())
         .on('data', (f: File) => splitFiles.set(unroot(f.path), f))
         .pipe(defaultProject.rejoinHtml())
         .on('data', (f: File) => joinedFiles.set(unroot(f.path), f))
         .on('end', () => {
-          const expectedSplitFiles = [
+          /* no spread operator in Node 4, so use apply... */
+          Array.prototype.push.apply(expectedSplitFiles, [
```
FredKSchott: `expectedSplitFiles` is always empty at this point, why use push here instead of just setting the array?

FredKSchott (changes requested):
Looks good, just a few questions.

Also, assigning @justinfagnani to review based on his feedback on the original PR.

src/polymer-project.ts
tony19: `splitScript` needs to use a `for`-loop because it uses `continue` at line 236 for unrecognized script types, while `splitStyle` doesn't. I could either (1) refactor this function so that the loop types are the same, or (2) restore the original `for`-loop in `splitStyle` for consistency (i.e., undo cc93d2a57c8f8335246d6d5e179eb4911fcaee37).

src/test/polymer-project_test.ts
tony19: To maintain its `const`ness. (The same is true for `expectedJoinedFiles[]`)

CHANGELOG.md
tony19: Ok, will do. I was anticipating your change in your PR for 0.7.0, which would result in a merge conflict here.

src/polymer-project.ts
FredKSchott: you can use `return` in the `forEach()` iterator to achieve the same behavior as `continue`

src/test/polymer-project_test.ts
FredKSchott: I like the idea, but this confuses more than it helps. It also doesn't help as much as you would hope, since the array can still be modified.

I'd rather see you use let to simplify, and then Object.freeze() to keep the array from being modified. Or use `setup()` to create the array for each test.

src/polymer-project.ts
tony19: Oh, didn't think of that 👍

src/test/polymer-project_test.ts
tony19: Ok, will do.

CHANGELOG.md
tony19: Done in c552144

src/polymer-project.ts
tony19: Done in 7f1e853

src/test/polymer-project_test.ts
tony19: Done in 02334e1

tony19:
Let me know if I should squash some of these commits.

FredKSchott (approved):
LGTM, but we can't merge until we can answer the same concerns that were brought up in https://github.com/Polymer/polymer-build/pull/40. Namely, that we still need to support splitting and not rejoining for CSP compliance (similar to https://github.com/PolymerLabs/crisper). But styles break if they aren't rejoined. So we need to add an option to rejoinHtml() to tell it to only rejoin CSS (and leave scripts split). 

@tony19 do you have any interest in tackling that? Otherwise I can try getting to it early next week.

tony19:
I'm interested, but not sure how soon I could get to it. Can that be a separate PR?

justinfagnani (changes requested):
Please wait for me to finish a review before merging. I still have to think about what it means to split CSS that's not legal externally. We need a way to ensure that `<custom-style>` and styles in templates actually work.

stramel:
Status?

cdata:
> We need a way to ensure that <custom-style> and styles in templates actually work.

This seems like something polymer-cli should ensure, but to venture a proposal: polymer-build should not care. polymer-build is the tool that assumes I know what I'm doing, no?

stramel:
Any update on this? I would love to see this get added in. @justinfagnani

abdonrd:
Any news?

stramel:
I know @rictic worked on updating the CSS parser in analyzer, https://github.com/Polymer/polymer-analyzer/pull/707, maybe he can weigh in on this?

enigma127:
Any news?

aomarks:
Closing this pull request because this repo is now archived. If this PR is still active, please submit a new one at https://github.com/Polymer/tools.

PR #108, opened 2017-01-14
Switch vulcanize out for polymer-bundler

usergenic:
Changed over all logic for bundling/sharding/etc to use what `polymer-bundler` provides and removed all references to `vulcanize` and `hydrolysis`.

Did not change any public API for `polymer-build` so this could be a seamless upgrade.  🤞 


src/analyzer.ts
```diff
@@ -485,29 +479,4 @@ export class StreamLoader implements BackwardsCompatibleUrlLoader {
     this.analyzer.pushDependency(urlPath);
     return waitForFile;
   }
-
```
justinfagnani: 🎉

src/bundle.ts
```diff
@@ -61,213 +61,36 @@ export class Bundler extends Transform {
...
-            reject(err);
-          } else {
-            resolve({
-              url: fragmentPath,
-              contents: doc,
-            });
+    const strategy = this.config.shell ?
+        generateShellMergeStrategy(this.config.shell, 2) :
+        undefined;
+    return this.bundler
```
justinfagnani: you can do: `const docCollection = await this.bundler.bundle(...);`

src/bundle.ts
```diff
@@ -61,213 +61,36 @@ export class Bundler extends Transform {
...
       done();
     });
   }
 
   async _buildBundles(): Promise<Map<string, string>> {
-    const bundles = await this._getBundles();
-    const sharedDepsBundle = (this.config.shell) ?
-        urlFromPath(this.config.root, this.config.shell) :
-        this.sharedBundleUrl;
-    const sharedDeps = bundles.get(sharedDepsBundle) || [];
```
justinfagnani: many 🎉 's here :)

web-padawan:
As of now, having shell configured, `shared_bundle_1.html` is still created instead of pushing shared imports to shell.

src/bundle.ts
```diff
@@ -44,6 +39,11 @@ export class Bundler extends Transform {
     this.config = config;
     this.analyzer = analyzer;
     this.sharedBundleUrl = 'shared-bundle.html';
+    this.bundler = new Bundler(<BundlerOptions>{
```
usergenic: These typings might be overkill. Was trying to please annoyance warnings from the tslint. . .

usergenic:
@web-padawan just fixed the shell merge scenario in a7605123ce9f78482f85e8dc701d35f755e15cd8, fyi.

usergenic:
@FredKSchott PTAL.  It's-a-working!

src/bundle.ts
```diff
@@ -61,213 +60,40 @@ export class Bundler extends Transform {
...
+    let strategy: BundleStrategy;
+    const fragments = Array.from(this.config.allFragments);
     if (this.config.shell) {
-      const shellFile = this.analyzer.getFile(this.config.shell);
-      console.assert(shellFile != null);
-      const newShellContent = this._addSharedImportsToShell(bundles);
-      shellFile.contents = new Buffer(newShellContent);
+      strategy = generateShellMergeStrategy(
+          urlFromPath(this.config.root, this.config.shell), 2);
+      fragments.push(this.config.shell);
```
FredKSchott: `this.config.allFragments` should already contain `this.config.shell` (lots of term overloading going on here, I know).

package.json
```diff
@@ -26,7 +26,6 @@
     "@types/vinyl": "^1.1.29",
     "@types/vinyl-fs": "0.0.28",
     "dom5": "^2.0.1",
-    "hydrolysis": "^1.23.3",
```
FredKSchott: 🎉

src/bundle.ts
```diff
@@ -61,213 +60,40 @@ export class Bundler extends Transform {
...
   _flush(done: (error?: any) => void) {
     this._buildBundles().then((bundles: Map<string, string>) => {
-      for (const fragment of this.config.allFragments) {
-        const file = this.analyzer.getFile(fragment);
-        console.assert(file != null);
-        const contents = bundles.get(fragment);
+      for (const filename of bundles.keys()) {
+        const filepath = pathFromUrl(this.config.root, filename);
+        let file =
+            this.analyzer.getFile(filepath) || new File({path: filepath});
```
FredKSchott: if the analyzer has never seen the fragment before, that's bad. Consider just not handling that case or throwing a special error here so that this isn't swallowed.

usergenic: There's no way for the analyzer to have seen a file generated by the bundler like "shared_bundle_1.html", which is what this case is for.

src/bundle.ts
```diff
@@ -61,213 +60,40 @@ export class Bundler extends Transform {
...
-
+    let strategy: BundleStrategy;
+    const fragments = Array.from(this.config.allFragments);
     if (this.config.shell) {
-      const shellFile = this.analyzer.getFile(this.config.shell);
-      console.assert(shellFile != null);
-      const newShellContent = this._addSharedImportsToShell(bundles);
-      shellFile.contents = new Buffer(newShellContent);
+      strategy = generateShellMergeStrategy(
+          urlFromPath(this.config.root, this.config.shell), 2);
```
web-padawan: Will you add an option to configure this magic number `2`?
I mean, there should be a way to try the benefit of generating more shared bundles.

src/bundle.ts
usergenic: There was extra dumb in that I didn't even use the fragments variable.  Fixed.  Thanks!

src/bundle.ts
usergenic: Good point.  This PR aims to eliminate or minimize changes to the build API and basically drop-in-replace polymer-bundler for vulcanize etc.  Adding configuration option here should be separate PR, but is probably worth looking at soon, but we want to be sure to do the best job of handling the bundling scenarios we are familiar with first.

usergenic:
Thought I already submitted this comment, but I can't find it so apologize if repost:

@FredKSchott I ran build on shop and starter kit and served the bundled output and they were happy.

I am going to try and correct https://github.com/Polymer/polymer-bundler/issues/412 behavior in bundler so this doesn't leak into the cli behavior.  Therefore, we can hold-off accepting this PR until new bundler release.

FredKSchott:
test (please ignore)

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 * `BuildAnalyzer.sources` & `BuildAnalyzer.dependencies` are now `BuildAnalyzer.sources()` & `BuildAnalyzer.dependencies()`, respectively. This change only affects uses who are importing and/or using the `BuildAnalyzer` class directly.
 * Fix issue where files were being loaded immediately, before the build stream was started.
-
+* Completed the migration away from `hydrolysis` to `polymer-analyzer` and `vulcanize` to `polymer-bundler`.
```
FredKSchott: test (please ignore)

abdonrd:
Yay! `v2.0.0-pre.3` is out!

usergenic:
(tests are all passing, but travis won't see until I can cut a new polymer-bundler release after https://github.com/polymer/polymer-bundler/pull/423 lands.

src/bundle.ts
```diff
@@ -44,15 +38,21 @@ export class Bundler extends Transform {
...
       file: File,
       _encoding: string,
       callback: (error?: any, data?: File) => void): void {
-    // If this file is a fragment, hold on to the file so that it's fully
-    // analyzed by the time down-stream transforms see it.
-    if (this.config.isFragment(file.path)) {
+    // If this file is a fragment or an entrypoint, hold on to the file so that
+    // it's fully analyzed by the time down-stream transforms see it.
+    if (this.config.isFragment(file.path) ||
+        !!this.config.entrypoint && this.config.entrypoint === file.path) {
```
FredKSchott: - `this.config.entrypoint` is always set
- Do we always want to bundle the entrypoint? We currently use this logic for deciding when the entrypoint is a fragment, can you confirm? https://github.com/Polymer/polymer-project-config/blob/master/src/index.ts#L225

src/bundle.ts
FredKSchott: ah, got it 👍

src/bundle.ts
```diff
@@ -61,213 +61,41 @@ export class Bundler extends Transform {
...
-    // Append all shared imports to the end of <head>
-    const head = dom5.query(doc, dom5.predicates.hasTagName('head'));
-    for (const dep of shellDeps) {
-      const newImport = dom5.constructors.element('link');
-      dom5.setAttribute(newImport, 'rel', 'import');
-      dom5.setAttribute(newImport, 'href', dep);
-      dom5.append(head, newImport);
+    const bundleEntrypoints = Array.from(this.config.allFragments);
+    if (!!this.config.entrypoint) {
+      bundleEntrypoints.push(this.config.entrypoint);
```
FredKSchott: similar to above, entrypoint is already a fragment when no shell or fragments are defined. This would double-push the entrypoint in.

usergenic: Removed.  There was a (terrible?) decision to start to use the term entrypoints in polymer-bundler to refer to what build and cli refer to as fragments.  This caused some confusion.  I've reverted the bit where I try to force index.html into the bundle set until we have a better handle on how to rewrite index.html's paths etc.

src/bundle.ts
usergenic: You're right-- bundling entrypoint problematic.  46b5c06 removes this mistake.

usergenic:
Okay @FredKSchott PTAL.  I've undid the bads.

abdonrd:
Yay! 🎉 New pre-release?

PR #111, opened 2017-01-17
add error handling to detect missing files

### LINKED ISSUE #88: Silent build failure when importing not-existing file
Occasionally forgotten import of file which has been deleted caused this for me: 

### Description
```
init_1   | [06:45:38] Starting 'vulcanize'...
init_1   | [06:45:50] The following tasks did not complete: default, vulcanize
init_1   | [06:45:50] Did you forget to signal async completion?
```

Gulp task looks like this:

```
const errorHandler = function (err) {
    // exit with a 'failure' code on build errors
    console.error(err.message);
    process.exit(1);
};

// Vulcanize components, generate shared file
function vulcanize () {
    return merge(project.sources(), project.dependencies())
        .pipe(project.analyzer)
        .on('error', errorHandler)
        .pipe(project.bundler)
        .on('error', errorHandler)
        .pipe(gulp.dest(TMP));
}
```

### Versions & Environment
- polymer-build: 0.5.1
- node: 6.9.1
- Operating System: Mac Os X El capitan

#### Expected Results
An error should be thrown (e. g. `Error: file ... does not exist`)

#### Actual Results
Build fails with no errors and 0 exit code.

FredKSchott:
Fixes #88

Currently, when a dependency is analyzed in your application but is missing from your file system, the build can hang. This happens because we delegate source-file loading to a vinyl stream, while we control dependency-file loading. In this system, the build hangs when a our analyzer assumes a source-file that it needs will be loaded by vinyl if that file doesn't actually exist. Mistyped import paths can frequently cause these errors.

This PR adds a check when the source stream completes to make sure that the analyzer never waits for source files that are never coming through the vinyl stream.

/cc @justinfagnani @usergenic


custom_typings/slash.d.ts
```diff
@@ -1,3 +1 @@
-declare module 'slash' {
-  export function slash(source: string): string;
-}
+
```
FredKSchott: doh, I thought I had. good catch

FredKSchott:
@justinfagnani I've pinged you for this one a few times now, assuming you're good with @usergenic's review here?

tony19:
@FredKSchott FYI, I just tested this branch on the originally reported error scenario in #88, and it still silently fails.

src/test/analyzer_test.ts
```diff
@@ -166,9 +166,59 @@ suite('Analyzer', () => {
...
+      () => {
+        const root = `test-fixtures/bad-src-import`;
+        const config = new ProjectConfig({
+          root: root,
+          entrypoint: 'index.html',
+          sources: ['src/**/*'],
+        });
+        const analyzer = new BuildAnalyzer(config);
+
+        return waitForAll([analyzer.sources(), analyzer.dependencies()])
```
justinfagnani: this test isn't testing the behavior you implemented directly - that you emit errors on particular streams, sometimes both. I think it'd be better to test the behavior directly by adding listeners to the streams. It'll make it clear what you're expecting.

tony19:
@FredKSchott @justinfagnani I'm going to create a GitHub project that isolates the problem for the repro. It's not as simple to reproduce as I had hoped for.

justinfagnani:
@tony19 thanks a lot for doing that!

tony19:
@FredKSchott @justinfagnani Here's the repro: https://github.com/tony19-sandbox/polymer-build-issue-88

It's almost as though `polymer-cli` (or `polymer-build`) isn't waiting properly for the file-not-found errors.

src/analyzer.ts
```diff
@@ -198,15 +199,29 @@ export class BuildAnalyzer {
     //   1. The resolver stream, to resolve each file loaded via the analyzer
     //   2. The analyzer stream, to analyze app fragments for dependencies
     this._sourcesProcessingStream =
-        this._sourcesStream.pipe(new ResolveTransform(this))
+        this._sourcesStream
+            .on('error',
+                (err: Error) =>
+                    this._sourcesProcessingStream.emit('error', err))
+            .pipe(new ResolveTransform(this))
+            .on('error',
```
FredKSchott: Not sure I completely understand your question, but if you're asking "doesn't this end up re-emitting a `_sourcesStream` error already caught on L203?" the answer is that `pipe()` doesn't propagate errors, so the `ResolveTransform` stream would never see or re-emit an error that occurred in `this_sources` stream. (See https://github.com/nodejs/node/blob/master/lib/stream.js#L26)

src/analyzer.ts
```diff
@@ -198,15 +199,29 @@ export class BuildAnalyzer {
...
     //   1. The vinyl loading stream, to load file objects from file paths
     //   2. The resolver stream, to resolve each loaded file for the analyzer
     this._dependenciesProcessingStream =
-        this._dependenciesStream.pipe(new VinylReaderTransform())
+        this._dependenciesStream
+            .on('error',
+                (err: Error) =>
+                    this._dependenciesProcessingStream.emit('error', err))
+            .pipe(new VinylReaderTransform())
+            .on('error',
```
FredKSchott: see response above

src/analyzer.ts
```diff
@@ -263,6 +278,32 @@ export class BuildAnalyzer {
...
+   * Perform some checks once we know that `_sourcesStream` is done loading.
+   */
+  private onSourcesStreamComplete() {
+    // Emit an error if there are missing source files still deferred. Otherwise
+    // this would cause the analyzer to hang.
+    if (this.loader.hasDeferredFiles()) {
+      for (const fileUrl of this.loader.deferredFiles.keys()) {
+        // TODO(fks) 01-13-2017: Replace with config.isSource() once released
+        if (minimatchAll(fileUrl, this.config.sources)) {
+          this.emitError(new Error(`Not found: ${fileUrl}`));
```
FredKSchott: good catch, we can actually do this for all "not found" errors

FredKSchott:
@tony19 glad you got the right version sorted out, but now I'm confused what is/isn't working at this point. Is build still broken in that repo for you? For your real project, can you find a way to repro the hanging build before we merge this?

tony19:
@FredKSchott I admit my earlier comment is a bit confusing even as I read it back now. ;) I'll edit it later.

To clarify, I confirmed that this PR resolves the case reported in https://github.com/Polymer/polymer-build/issues/88#issuecomment-272225906, where there's only one "bad" import. However, I found a different scenario in my real project that still fails with this PR, and I've isolated the problem [here](https://github.com/tony19-sandbox/polymer-build-issue-88). I discovered that deleting seemingly unrelated files, such as PNGs, would cause this PR to correctly spit out "file-not-found" errors for bad imports. Otherwise, it silently fails.

FredKSchott:
@tony19 I can't reproduce a silent build on either `0.18.0-alpha.8`or our latest build branch of the CLI (with this PR branch npm linked into the CLI). For both, I get the following output: 

```
warn:    "sourceGlobs" config option has been renamed to "sources" and will no longer be supported in future versions
warn:    "includeDependencies" config option has been renamed to "extraDependencies" and will no longer be supported in future versions
info:    Preparing build...
info:    Building application...
info:    Generating build/unbundled...
info:    Generating build/bundled...
error:   Uncaught exception: Error: ENOENT: no such file or directory, open '/Users/fschott/Code/polymer-build-issue-88/bower_components/neon-animation/neon-animated-pages.html'
error:   Error: ENOENT: no such file or directory, open '/Users/fschott/Code/polymer-build-issue-88/bower_components/neon-animation/neon-animated-pages.html'
    at Error (native)
```

tony19:
@FredKSchott That's weird. It was happening 100% of the time for me, but after putting it up on Travis, it was behaving correctly there. 🤔 

... *hours of trial and error later* 😄  ...

I've finally created a more reliable repro that also fails on [Travis](https://travis-ci.org/tony19-sandbox/polymer-build-issue-88/jobs/196086013). Can you try the update?

FredKSchott:
@tony19 I still can't reproduce. Can you make travis run with the `--verbose` flag? 

This also might be a CLI problem and not a build problem. Could you run this test with a polymer-build in a gulpfile instead, like in https://github.com/PolymerElements/generator-polymer-init-custom-build/blob/master/generators/app/gulpfile.js ?

tony19:
@FredKSchott Ok, I've enabled `--verbose` and added the `gulpfile.js`, which also silently fails in [Travis](https://travis-ci.org/tony19-sandbox/polymer-build-issue-88/jobs/196493192). The Gulp build aborts at `Analyzing build dependencies...`.

FredKSchott:
@justinfagnani PTAL, just some small changes since our last talk

@tony19 I still can't reproduce on my machine, and going back and forth on your travis setup isn't going to be a good use of time. Lets get this merged and then try to come up with a better repro case that works on all of our machines. I'm fairly confident that this branch isn't introducing any new bugs related to hanging builds, and it still nice solves all the other ones we've seen. 

Once this is merged you can just run `npm install polymer-build@next` on travis which should cut down on your test setup complexity a bit as well.

src/analyzer.ts
justinfagnani: I told you in person already, but that sucks :/ One day the world will be better...

PR #118, opened 2017-01-31
pull html splitting/rejoining logic into new class

### LINKED ISSUE #113: splitHtml() and rejoinHtml() cannot be run in parallel
Because they share a single state, build streams cannot be forked and run in parallel if they both include splitting and rejoining streams.

FredKSchott:

This pulls `splitHtml()` and `rejoinHtml()` off of Project and into a new `HTMLSplitter` class, so that multiple split/rejoin states can be managed seperately. This fixes #113. It adds no new logic to control splitting & rejoining but sets the stage so that it can be added in a non-breaking way (todo later).


src/html-splitter.ts
```diff
@@ -0,0 +1,258 @@
...
+const extensionsForType: {[mimetype: string]: string} = {
+  'text/ecmascript-6': 'js',
+  'application/javascript': 'js',
+  'text/javascript': 'js',
+  'application/x-typescript': 'ts',
+  'text/x-typescript': 'ts',
+};
+
+/**
+ * HTMLSplitter represents the shared state of files as they are passed through
```
justinfagnani: For a user I don't think they care/need to know about shared state - or how this works. Let's just say what it does:

An HtmlSplitter extracts inline scripts and styles from HTML into separate files so they can be individually processed by other transforms, like compilers or minifiers, and optionally rejoins them back into their original HTML files.

You must use a separate instance of HtmlSplitter per stream, do not reuse one HtmlSplitter's split and rejoin transforms on multiple streams.

src/html-splitter.ts
```diff
@@ -0,0 +1,258 @@
...
+export class HtmlSplitter {
+  private _splitFiles: Map<string, SplitFile> = new Map();
+  private _parts: Map<string, SplitFile> = new Map();
+
+  /**
+   * Returns a new `Transform` stream that splits inline script and styles into
+   * new, separate files that are passed out of the stream.
+   */
+  split(): Transform {
+    return new HtmlSplitTransform(this);
```
justinfagnani: I think this and rejoin should be simple properties/getters now - we only want instance one per HtmlSplitter instance.

src/html-splitter.ts
```diff
@@ -0,0 +1,258 @@
...
+
+  getParentFile(childPath: string): SplitFile {
+    return this._parts.get(childPath);
+  }
+}
+
+/**
+ * Represents a file that is split into multiple files.
+ */
+export class SplitFile {
```
justinfagnani: Does anyone use this export?

src/html-splitter.ts
FredKSchott: Code comments are for our developers, not our consumers, right? This is explained for the user in the README.

src/html-splitter.ts
FredKSchott: Even given this plan? https://docs.google.com/document/d/1XWQgtALNVQTnPCuLAevNv8ihZEDgH83YqvYZg7F0RTc/edit

I agree with the benefit of getters to enforce a single splitter per instance, but not sure it's worth it if it introduces a future, avoidable breaking interface change. Plus there are other ways to enforce single instances while keeping our planned function interface.

src/html-splitter.ts
FredKSchott: not anymore! Removing...

src/html-splitter.ts
FredKSchott: Will merge with this current interface since it will more easily adapt to our future design for splitting/rejoining options without breaking. Will also add an issue to track an enhancement to prevent multiple splitting/rejoining instances from being created, which we can prioritize as needed.

FredKSchott: Issue created here: https://github.com/Polymer/polymer-build/issues/129

PR #125, opened 2017-02-07
remove duplicate merge-stream dependency

FredKSchott:


package.json
```diff
@@ -52,7 +51,7 @@
     "gulp-tslint": "^6.1.1",
     "gulp-typescript": "^2.13.4",
     "gulp-typings": "^2.0.0",
-    "merge-stream": "^1.0.0",
+    "merge-stream": "^1.0.1",
```
justinfagnani: Does this need to be 1.0.1? Prefer the lowest bounds that works to avoid duplicates in npm

FredKSchott: yea it does, there was an important fix released in `v1.0.1` to a bug that otherwise hides errors in our users build pipelines.

PR #144, opened 2017-03-01
Bundle from stream

### LINKED ISSUE #110: Bundler causes rejoining of original streams instead of modified ones
### Description
When the build stream is passed through the bundler, the result contains the original split streams instead of the modified ones (e.g., processed by `uglify` or other minifiers).

### Versions & Environment
- polymer-build: 0.6.0
- node: 7.4.0
- Operating System: macOS Sierra 10.12

#### Steps to Reproduce

 1. Generate project from [`generator-polymer-init-custom-build` 2.0.0](https://github.com/PolymerElements/generator-polymer-init-custom-build/releases/tag/v2.0.0).
 
 2. Install `gulp-uglify`:

        yarn add -D gulp-uglify

 3. Edit the project's `gulpfile.js` to use `gulp-uglify`, and uncomment [`gulpfile.js:46`](https://github.com/PolymerElements/generator-polymer-init-custom-build/blob/bb033e5/generators/app/gulpfile.js#L46), which runs `uglify`:

        const uglify = require('gulp-uglify');
        ...

        function build() {
          ...
          let sourcesStream = polymerProject.sources()
            .pipe(gulpif(/\.js$/, uglify()))

 4. Run `gulp build`.
 5. Observe the output of `build/src/my-view1.html` does **not** contain minified JS.
 6. Comment out [`gulpfile.js:69`](https://github.com/PolymerElements/generator-polymer-init-custom-build/blob/bb033e5/generators/app/gulpfile.js#L69), which pipes the build through the bundler:

        buildStream = buildStream.pipe(polymerProject.bundler);

 7. Run `gulp build`.
 8. Observe the output of `build/src/my-view1.html` contains minified JS.

#### Expected Results
Output files contain modified streams.

For example, minified JS like this:

```html
<script>Polymer({is:"my-view1"});let foo=1;console.log(foo++);</script>
```

#### Actual Results
Output files contain original/unmodified streams from `HTMLSplitter`.

For example:

```html
<script>
  Polymer({
    is: 'my-view1',
  });

  let foo = 1;
  console.log(foo++);
</script>
```

usergenic:
Fixes #110 by piping the stream into the Bundler through a generic File stream-based `UrlLoader`.
Note: This PR is dependent on features in https://github.com/Polymer/polymer-bundler/pull/439 and https://github.com/Polymer/polymer-bundler/pull/443 to be reviewed/merged and released.


src/bundle.ts
```diff
@@ -32,14 +36,22 @@ export class BuildBundler extends Transform {
...
     this.analyzer = analyzer;
     this.sharedBundleUrl = 'shared-bundle.html';
     this.bundler = new Bundler({
-      analyzer: analyzer.analyzer,
+
+      // TODO(usergenic): Creating a new Analyzer with a blank cache is going
+      // to mean, at least, a doubling of analysis efforts for bundling phase.
+      // Ideally we would fork existing analyzer and replace its urlLoader if
+      // there were an affordance to do so.
+      analyzer: new Analyzer({urlLoader: new StreamLoader(this)}),
```
rictic: use the newly released _fork()?

Or maybe, because it's an underscore method and so kinda experimental, it would be better to do something like

```typescript
let localAnalyzer;
if (analyzer._fork) {
  localAnalyzer = analyzer._fork();
  // do something to modify the url loader
} else {
  localAnalyzer = new Analyzer({/*...*/});
}

File: src/bundle.ts [hunk 2]
```diff
@@ -95,3 +108,52 @@ export class BuildBundler extends Transform {
...
+
+  canLoad(url: string): boolean {
+    return this.bundler.analyzer.analyzer.canResolveUrl(url);
+  }
+
+  getFile(url: string): File {
+    return this.bundler.files.get(url);
+  }
+
+  load(url: string): Promise<string> {
```
rictic: I'd like to propose a convention that all Promise-returning functions should be async.

Not only do you get `await` goodness, but you can also be confident that any thrown error becomes a rejection. This is more predictable than splitting the failure codepath, with some becoming rejections and some being thrown up the call stack.

src/bundle.ts
```diff
@@ -95,3 +108,52 @@ export class BuildBundler extends Transform {
...
   }
 }
+
+export type ResolveFileCallback = (a: string) => void;
+export type RejectFileCallback = (err: Error) => void;
+export type DeferredFileCallbacks = {
+  resolve: ResolveFileCallback; reject: RejectFileCallback;
+};
+
+class StreamLoader implements UrlLoader {
```
rictic: This looks very similar to the class of the same name in src/analyzer.ts, what's up with the difference?

Document it in source?

src/bundle.ts
```diff
@@ -95,3 +108,52 @@ export class BuildBundler extends Transform {
...
+  bundler: BuildBundler;
+
+  // Store files that have not yet entered the Analyzer stream here.
+  // Later, when the file is seen, the DeferredFileCallback can be
+  // called with the file contents to resolve its loading.
+  deferredFiles = new Map<string, DeferredFileCallbacks>();
```
rictic: Add a comment that mentions whose responsibility it is to go through these and resolve or reject them.

src/bundle.ts
usergenic: I like the idea of using `_fork()` except that its still to early to rely on in this PR.

 - I don't have a good solution for `// do something to modify the url loader` yet without a public API change to `_fork()` or deep reaching into Analyzer's privates.
 - There needs to be some code to identify the document paths to invalidate based on which files have changed since original analysis or the cache will preserve the state of the original scanned scripts/stylesheets instead of getting the downstream (minified etc) versions during inlining.

usergenic: I'm not 100% where that `file-changed?` logic should live yet, but I think `_fork()` may need API change in `Analyzer` and `AnalysisContext` to support.

- Added https://github.com/Polymer/polymer-analyzer/issues/547 "Extend the `Analyzer#_fork()` API to support replacing the `urlLoader` option (possibly replace any of the initialization options?)"
- Added https://github.com/Polymer/polymer-analyzer/issues/548 "Can Analyzer/AnalysisCache analyze in a mode where it tests for file content changes?"

src/bundle.ts
usergenic: I decided to change this into a generic FileMapUrlLoader (now a different file and not married to a BuildBundler instance.)

src/bundle.ts
usergenic: This was a dumb cargo'd item that is totally not used and I have deleted it.

src/bundle.ts
usergenic: 100% yes.

usergenic:
@rictic PTAL - feedback addressed :D

src/bundle.ts
rictic: :+1:

src/file-map-url-loader.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import File = require('vinyl');
+import {UrlLoader} from 'polymer-analyzer/lib/url-loader/url-loader';
+import {parseUrl} from 'polymer-analyzer/lib/utils';
+import {pathFromUrl} from './path-transformers';
+
+/**
+ * A map of paths to Files.
+ */
+export type FileMap = Map<string, File>;
```
usergenic: That sounds reasonable.  Done.

src/file-map-url-loader.ts
```diff
@@ -0,0 +1,58 @@
...
+    return true;
+  }
+
+  async load(url: string): Promise<string> {
+    const file = this.files.get(this._filePath(url))!;
+
+    if (file) {
+      return Promise.resolve(file.contents.toString());
+    } else {
+      return Promise.reject(undefined);
```
usergenic: Because I didn't think of it.  But now I will do it, because its better.  And I will try to remember that I can do that in `async` methods.

src/file-map-url-loader.ts
```diff
@@ -0,0 +1,58 @@
...
+  // We can always return true because we're just reading paths off a map.
+  canLoad(_url: string): boolean {
+    return true;
+  }
+
+  async load(url: string): Promise<string> {
+    const file = this.files.get(this._filePath(url))!;
+
+    if (file) {
+      return Promise.resolve(file.contents.toString());
```
usergenic: I was... just testing you.  You passed.  Done.

src/file-map-url-loader.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  // We can always return true because we're just reading paths off a map.
+  canLoad(_url: string): boolean {
+    return true;
+  }
+
+  async load(url: string): Promise<string> {
+    const file = this.files.get(this._filePath(url))!;
+
+    if (file) {
```
usergenic: Oh really?  Didn't think it would because Object but totes switching to null check.  Done.

src/bundle.ts
```diff
@@ -32,14 +33,23 @@ export class BuildBundler extends Transform {
   bundler: Bundler;
   sharedFile: File;
 
+  files = new Map<string, File>();
+
   constructor(config: ProjectConfig, analyzer: BuildAnalyzer) {
-    super({objectMode: true});
+    super({objectMode: true, highWaterMark: 10000});
```
usergenic: Removed this setting.  I misunderstood its example in the BuildAnalyzer Transform.

usergenic:
Thanks for the feedback again, @rictic.  PTAL!

src/file-map-url-loader.ts
rictic: Oh, sorry, you're totes right. I was thinking of `file` as a string, not a File. My bad!

src/analyzer.ts
```diff
@@ -523,20 +512,17 @@ export class StreamLoader implements UrlLoader {
...
 
   load(url: string): Promise<string> {
     logger.debug(`loading: ${url}`);
     const urlObject = parseUrl(url);
 
-    // Resolve external files as empty strings. We filter these out later
-    // in the analysis process to make sure they aren't included in the build.
-    if (isDependencyExternal(url)) {
-      return Promise.resolve('');
+    if (!this.analyzer.analyzer.canResolveUrl(url)) {
```
web-padawan: Why not to use `canLoad` here? Wouldn't this be more DRY?

src/bundle.ts
```diff
@@ -32,14 +33,23 @@ export class BuildBundler extends Transform {
   bundler: Bundler;
   sharedFile: File;
```
rictic: As this class is exported, how much of this API can we make private and/or readonly?

Also, would it be a breaking change to rename `analyzer` to `buildAnalyzer`, as this class also does use the Analyzer class

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 <!-- List New Changes Here -->
+* Build now delegates authority to the Analyzer for what urls are package external instead of local heuristics.
+* Bundling now processes files coming in from streams, to support things like js-minification before bundling.
```
rictic: Does this PR also mean that we'll be able to print warnings with source ranges in `polymer build`? Is there a bug to track this?

It'll be a huge win once we can. Just this morning someone reported an error in their build that – I think – would have been obvious with a source range but was pretty opaque without.

src/analyzer.ts
```diff
@@ -523,20 +512,17 @@ export class StreamLoader implements UrlLoader {
...
-  canLoad(_url: string): boolean {
-    // We want to return true for all files. Even external files, so that we
-    // can resolve them as empty strings for now.
-    return true;
+  // We can't load external dependencies.
+  canLoad(url: string): boolean {
+    return this.analyzer.analyzer.canResolveUrl(url);
   }
 
   load(url: string): Promise<string> {
```
rictic: Make this async?

src/analyzer.ts
```diff
@@ -523,20 +512,17 @@ export class StreamLoader implements UrlLoader {
...
     logger.debug(`loading: ${url}`);
     const urlObject = parseUrl(url);
 
-    // Resolve external files as empty strings. We filter these out later
-    // in the analysis process to make sure they aren't included in the build.
-    if (isDependencyExternal(url)) {
-      return Promise.resolve('');
+    if (!this.analyzer.analyzer.canResolveUrl(url)) {
+      return Promise.resolve(undefined);
```
rictic: Hm, we don't have strict null checks turned on in this project yet?

src/analyzer.ts
usergenic: Good catch.  DRY'd :)

src/analyzer.ts
usergenic: Converted this to a `throw new Error`

src/analyzer.ts
usergenic: Yeah!

src/bundle.ts
usergenic: Okay did lots of that.  Privatized what I thought made sense, renamed properties to comport to their classes for clarity and also removed a totally unused thing called `sharedBundleUrl` that didn't actually drive/inform anything correctly and made service workers attempt to precache file that didn't exist.

CHANGELOG.md
usergenic: I don't see a specific bug for this.  It feels like we might need that behavior in both `polymer-build` AND `polymer-bundler`.  I'll add bugs to both.

usergenic: https://github.com/Polymer/polymer-bundler/issues/450 + https://github.com/Polymer/polymer-build/issues/148

stramel:
Thanks @usergenic 🎉

PR #157, opened 2017-03-10
Change format script in package.json to not reference non-existant test/ directory (keeps printing warn message)

usergenic:
`"format": "find src/ test/ -iname '*.ts' -o -iname '*.js' | xargs clang-format --style=file -i"` is now `"format": "find src/ -iname '*.ts' -o -iname '*.js' | xargs clang-format --style=file -i"`


usergenic:
(thanks!)

PR #161, opened 2017-03-15
Fix paths for windows

usergenic:
- Fixed some tests that were breaking because assumptions in the test were they were run in posix environment.  Used path.join instead of '/' in strings.
 - Fixed the service-worker.ts code to use a posixified path string when providing a prefix-to-replace since sw-precache apparently converts all paths to that form.
 - This fixes all known Windows path issues with respect to BuildBundler and possibly some unknown issues.


src/path-transformers.ts
```diff
@@ -44,6 +44,10 @@
 
 import * as path from 'path';
 
+export function posixifyPath(filepath: string): string {
```
FredKSchott: Can you use this in `urlFromPath()` as well?

src/service-worker.ts
```diff
@@ -110,8 +111,10 @@ export async function generateServiceWorker(options: AddServiceWorkerOptions):
     return path.join(buildRoot, filePath);
   });
 
-  // swPrecache will determine the right urls by stripping buildRoot
-  swPrecacheConfig.stripPrefix = buildRoot;
+  // swPrecache will determine the right urls by stripping buildRoot.
+  // NOTE:(usergenic) sw-precache generate() apparently to replaces the
```
FredKSchott: ~~to~~ replaces

FredKSchott: Also wat??? That's crazy. Can you also file a bug with them to get this fixed? (At the very least, this means that some of their README examples don't actually work on Windows).

src/path-transformers.ts
usergenic: Done.  I went and made some additional simplifications to the path-transformers file.

src/service-worker.ts
usergenic: I fixed the comment.  Thanks.

I'm not sure technically this counts as a *bug* with sw-precache, and of course if they fix it, it will break for people who are working around it by using forward slashes anyway.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 <!--## Unreleased-->
+* Fixed issue with Service Worker generation in Windows environment where full paths were put into its precacheConfig instead of relative paths.
+* Bundled files always use canonical platform separators in their paths now.  Sometimes files would have back-slashes and sometimes forward-slashes on Windows, depending on how they arrived at the Bundler and this caused the Analyzer to treat some files as missing.
```
rictic: Consider:

Sometimes files -> Previously, sometimes files

src/path-transformers.ts
```diff
@@ -44,34 +44,41 @@
...
-  const isPlatformWin = /^win/.test(process.platform);
-  let filepath: string;
+export function posixifyPath(filepath: string): string {
+  if (isPlatformWindows()) {
+    // Strip "extended-length path" prefix.
+    filepath = filepath.replace(/^\\\\\?\\/, '');
+    // Replace all \ with /
+    filepath = filepath.replace(/\\/g, '/');
+  }
+  return filepath;
```
rictic: These are subtle operations with a number of weird edge cases, is there a more standard library we could use?

src/path-transformers.ts
```diff
@@ -44,34 +44,41 @@
...
+      path.posix.join(posixifyPath(root), path.posix.join('/', url))));
+}
 
-  // Otherwise, just return the relative path between the two
-  return path.relative(root, filepath);
+export function platformifyPath(filepath: string): string {
+  // Replaces all / with \ on win32.  Otherwise this is a noop.
+  // TODO(usergenic): Should we produce an "extended-length path" in win32 if
+  // the path length is over 259 on win32?
+  return filepath.replace(/\//g, path.sep);
```
rictic: Why not `path.normalize(filepath)`?

```javascript
> require('path').win32.normalize('./foo/bar/baz')
'foo\\bar\\baz'
```

It doesn't seem like it damages paths in any way that we'd be harmed by for any of the cases I tested:

```
   ./foo         → foo
   ../foo        → ../foo
   ../foo/       → ../foo/
   ../foo/../bar → ../bar
```

src/path-transformers.ts
```diff
@@ -44,34 +44,41 @@
...
+  return platformifyPath(decodeURI(
+      path.posix.join(posixifyPath(root), path.posix.join('/', url))));
+}
 
-  // Otherwise, just return the relative path between the two
-  return path.relative(root, filepath);
+export function platformifyPath(filepath: string): string {
+  // Replaces all / with \ on win32.  Otherwise this is a noop.
+  // TODO(usergenic): Should we produce an "extended-length path" in win32 if
+  // the path length is over 259 on win32?
```
rictic: I don't think so, it looks like the node fs module at least calls path._makeLong, which will make paths into windows extended-length paths

src/path-transformers.ts
rictic: I did some research, https://www.npmjs.com/package/upath looks like a reasonable option, though it has kinda a kitchen sink of functionality, and I'm not sure we need it. Recommend adding a comment to this effect for future-us

src/path-transformers.ts
```diff
@@ -44,34 +44,41 @@
...
+
+  if (!filepath.startsWith(root)) {
+    throw new Error(`file path is not in root: ${filepath} (${root})`);
   }
 
-  // Otherwise, just return the relative path between the two
-  return path.join(root, url);
+  // The goal is a relative URL from the root, so strip out the root and the
+  // leading slash, so '/my-app/subfolder/file.html' => 'subfolder/file.html'
+  return encodeURI(filepath.replace(root, '').replace(/^\//, ''));
```
rictic: path.relative? it'll also automatically posixify

src/path-transformers.ts
```diff
@@ -44,34 +44,41 @@
...
+  }
+  return filepath;
+}
 
-  // On windows systems, convert URL to filesystem path by replacing slashes
-  if (isPlatformWin) {
-    filepath = url.replace(/\//g, '\\');
-  } else {
-    filepath = url;
+export function urlFromPath(root: string, filepath: string) {
```
rictic: Please document this method and its intended use a bit more

src/test/path-transformers_test.ts
```diff
@@ -0,0 +1,71 @@
...
+  test('creates a filesystem path using the platform separators', () => {
+    const otherSeparator = pathSeparator === '/' ? '\\' : '/';
+    const path = pathFromUrl(RootPath, '/some/url/pathname');
+    assert.include(path, pathSeparator);
+    assert.notInclude(path, otherSeparator);
+  });
+
+  test('returns a path if url is absolute', () => {
+    const path = pathFromUrl(RootPath, '/absolute/path');
+    assert.equal(path, pathJoin(RootPath, 'absolute', 'path'));
```
rictic: It kinda seems like this is duplicating work that the Analyzer's resolver and loader should handle. We should strive as much as possible to do url -> fs path conversion in configurable places as in the resolvers and loaders, as this is something that clients, especially big ones, need to customize.

For one example, `../paper-button/paper-button.html` in a default app needs to end up pointing to `rootDir/bower_components/paper-button/paper-button.html`

I don't have a global view of this, but it doesn't seem like we're maintaining all the info needed to do the right thing here.

src/test/path-transformers_test.ts
```diff
@@ -0,0 +1,71 @@
...
+  test('will not go outside the root path', () => {
+    const path = pathFromUrl(RootPath, '../../../still/../root/path');
+    assert.equal(path, pathJoin(RootPath, 'root', 'path'));
+  });
+});
+
+suite('urlFromPath()', () => {
+
+  test('throws error when path is not in root', () => {
+    assert.throws(function() {
```
rictic: Optional: arrow function?

CHANGELOG.md
usergenic: Updated.

src/test/path-transformers_test.ts
usergenic: So there are a few places in Build where we use `pathFromUrl` to generate the filesystem path where something will be written to, because right now the Analyzer doesn't have a symmetrical parse/production interface for paths<->urls; urls are resolved and loaded by Analyzer and there's no means by which Analyzer can be given a url it's never seen, for example, and decide what the filesystem path should be.

On your example, I need to understand if you're saying there is a legitimate case for `../paper-button/paper-button.html` to ever become `rootDir/bower_components/paper-button/paper-button.html` when the base url is not `rootDir/bower_components/{anything}` in a build scenario.

rictic: Chatted it out in person. Doing bidirectional mapping between document urls and loader urls is still a fuzzy idea. I think it's something that we'll need to do, but we need to do some more exploration and requirements gathering first.

src/path-transformers.ts
```diff
@@ -85,6 +96,9 @@ export function urlFromPath(root: string, target: string): string {
 
   const relativePath = path.posix.relative(root, target);
 
+  // The startsWith(root) check is important on Windows because of the case
+  // where paths have different drive letters.  The startsWith('../') will
+  // catch the general not-in-root case.
```
rictic: Would it be clearer to catch the windows issue here by asserting that `relativePath` isn't absolute?

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 <!--## Unreleased-->
 * Fixed issue with Service Worker generation in Windows environment where full paths were put into its precacheConfig instead of relative paths.
-* Bundled files always use canonical platform separators in their paths now.  Sometimes files would have back-slashes and sometimes forward-slashes on Windows, depending on how they arrived at the Bundler and this caused the Analyzer to treat some files as missing.
+* Bundled files always use canonical platform separators in their paths now.  Previously, files might have either back-slashes or forward-slashes on Windows, depending on how they arrived at the Bundler and this caused the Analyzer to treat files as missing when mapping them by path.
```
rictic: +1 much easier to read!

PR #170, opened 2017-03-30
Fix issues with inlined files output by BuildBundler stream

usergenic:
 - Fixed https://github.com/Polymer/polymer-build/issues/167 where two copies of entrypoint files with same name are emitted by bundler stream: an un-bundled one, followed by bundled one.  
 - Fixed https://github.com/Polymer/polymer-build/issues/142 where html imports were emitted by bundler as individual files even though they were bundled.
 - This depends on a new release of polymer-bundler following review/merge of https://github.com/Polymer/polymer-bundler/pull/464

### DIFF
--- a/src/bundle.ts
+++ b/src/bundle.ts
@@ -30,6 +30,10 @@ export class BuildBundler extends Transform {
   private _buildAnalyzer: BuildAnalyzer;
   private _bundler: Bundler;
 
+  // A map of urls to file objects.  As the transform stream handleds files
+  // coming into the stream, it collects all files here.  After bundlling,
+  // we remove files from this set that have been inlined and replace
+  // entrypoint/fragment files with bundled versions.
   files = new Map<string, File>();
 
   constructor(config: ProjectConfig, buildAnalyzer: BuildAnalyzer) {
--- a/src/bundle.ts
+++ b/src/bundle.ts
@@ -77,9 +77,7 @@ export class BuildBundler extends Transform {
       });
       this._mapFile(file);
     }
-    for (const file of this.files.values()) {
-      this.push(file);
-    }
+    this.files.forEach((file) => this.push(file));
     // end the stream
     done();
   }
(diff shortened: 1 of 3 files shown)


justinfagnani:
@usergenic can you update to pull in Polymer/polymer-bundler#464 ?

usergenic:
Updated and ready for review!  PTAL.

/cc @azakus this is the fix you're looking for wrt duplicates.

usergenic:
(just pushed `--force` after rebase because master was so far ahead.)

azakus:
@usergenic yep, that's about what I did on my side, so it _should_ fix my issue. Thanks!

src/bundle.ts
```diff
@@ -55,18 +55,26 @@ export class BuildBundler extends Transform {
...
       const filepath = pathFromUrl(this.config.root, filename);
-      this.push(new File({
+      const file = new File({
         path: filepath,
         contents: new Buffer(bundles.get(filename)),
-      }));
+      });
+      this._mapFile(file);
+    }
+    for (const filepath of this.files.keys()) {
```
FredKSchott: you can just do this instead:
```

for (const file of myMap.values()) {
  this.push(file);
}
```

or even:

```
this.files.forEach((file) => this.push(file));
```

src/bundle.ts
```diff
@@ -82,11 +90,23 @@ export class BuildBundler extends Transform {
...
       contentsMap.set(
           bundleName, parse5.serialize(docCollection.get(bundleName).ast));
     }
     return contentsMap;
   }
+
+  _unmapBundledFiles(manifest: BundleManifest) {
+    for (const bundle of manifest.bundles.values()) {
+      for (const filename of bundle.files) {
+        this.files.delete(filename);
```
FredKSchott: I can't comment on the `this.files` definition above, but please document what `this.files` state represents.  In all other streams it represents a cache of all files, but now it's contents actually dictate the stream output. This would be useful to explain somewhere for later readers.

src/bundle.ts
usergenic: Great suggestion!  Done.

src/bundle.ts
usergenic: Added documentation for property.

PR #171, opened 2017-03-31
print warning with line and column info

baocang:
fixed https://github.com/Polymer/polymer-cli/issues/631

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * [BREAKING] Dropped support for node v4, added support for node v8. See our [node version support policy](https://www.polymer-project.org/2.0/docs/tools/node-support) for details.
 * Dependency updates.  Upgraded to new `polymer-bundler`, `polymer-analyzer` and `dom5` versions.
 * Fixed bug where `<html>`, `<head>` and `<body>` were added to documents mutated by `HtmlSplitter` and the `CustomElementsES5AdapterInjector`.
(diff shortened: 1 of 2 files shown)


rictic:
A bit of background first. Ideally we'd like to use the WarningPrinter from the analyzer, to get output like the linter gives, i.e.

```
          <link rel="import" href="./does-not-exist.html">
                                   ~~~~~~~~~~~~~~~~~~~~~
```

When we tried this last time though, we ran into trouble because we were analyzing some intermediate state, so the line numbers and column numbers didn't line up.

Getting better output here though would be a huge win, and I know @usergenic and @FredKSchott were making progress on that, so that we could use the WarningPrinter.

rictic:
Have you tried this out in your build? Does this draw the squiggles in the right places?

baocang:
@rictic I tried over our company project. and i also create a new project to test that.  the squiggles in the right places but with more than one warning in single line, it draw the squiggles only the last one. I think this is not a big problem for use. Thanks for you did the [WarningPrinter](https://github.com/Polymer/polymer-analyzer/commits/master/src/warning/warning-printer.ts)

is the line numbers and column numbers didn't line up has fixed?

image

59283d4f-aaae-43fa-9e9b-1e92aa2a486e

FredKSchott:
Now that analysis is happening before build, I think we actually are able to just use the analyzer's WarningPrinter

rictic:
This looks good to me, please update the CHANGELOG. I'd recommend something like

```
 * Print build-time warnings and errors with full location information, and precise underlines of the place where the problem was identified.
```

rictic:
Thanks for the PR!

PR #172, opened 2017-04-05
Add ability to generate push manifests

FredKSchott:
## Open Questions:

- [x] Is the expected behavior above sound? I talked a bunch with @keanulee about news & shop pushing and this seemed to maximize good pushes with minimum unnecessary pushes. 
- [x] I'm unfamiliar with how important the "as=" attribute on the link header, is it okay to match the push-manifest "type" to the file extension, or does this need to match some attribute in the HTML link/script element? If it's missing, is it okay for the push-manifest "type" property to be undefined?
- [x] @usergenic: In the app-shell bundled case, why is a shell import added to each fragment if the fragment is always loaded by the shell? This currently breaks the push-manifest in the bundled case. A somewhat hacky workaround to remove the shell could be added, but I wanted to confirm that this was expected before looking further.

---

- This PR was based off the work started by @slightlyoff in [push-manifest](https://github.com/slightlyoff/polymer-build/tree/push-manifest) branch.
- The repo/documentation for the "push manifest" schema can be found here: https://github.com/GoogleChrome/http2-push-manifest


keanulee:
@FredKSchott The `as` value is important - if set improperly, the browser may not match a requested resource with a pushed resource (you'll see two entries for the same file in the network panel and a console warning for a unused pushed resource). HTML imports should be `as=document`

src/push-manifest.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+/**
+ * Gets a link type from a file URL. Useful for generating a PushManifestEntry
+ * and needed for the "as=" value on Link headers.
+ */
+function
+getLinkTypeFromUrl(url: string):
+    LinkType|undefined {
+      switch (path.extname(url)) {
```
ebidel: Here's the list I used: https://github.com/GoogleChrome/http2-push-manifest/blob/master/lib/manifest.js#L30-L41

src/push-manifest.ts
```diff
@@ -0,0 +1,157 @@
...
+        pushManifest[relativeFragmentUrl] =
+            generatePushManifestEntries(Array.from(fragmentImportUrls));
+      }
+
+      return pushManifest;
+    }
+
+/**
+ * Returns a promise that resolves when a service worker has been generated
+ * and written to the build directory. This uses generateServiceWorker() to
```
ebidel: Think this doc string needs updating :)

FredKSchott:
@keanulee thanks for the extra info, that's helpful. Is this documented anywhere? I tried looking everywhere and couldn't find any reference to the `as` value, even in the official spec (it's not the easiest to search for).

Given what you've said though, I think that supports setting the "as" value automatically based on the file extension. I can't see a reason the user would ever want to manually set that value in their HTML to anything other than the file type, especially if that would result in double-pushed resources.

src/push-manifest.ts
keanulee: @ebidel Interesting how `.json` maps to `script` - when I was writing my codelab, Chrome wouldn't claim my json file if it was `as=script`, but it would if I omit the `as=` part completely https://github.com/PolymerLabs/prpl-ce-firebase/blob/master/final/firebase.json#L15

ebidel: Yea, looking at https://w3c.github.io/preload/#link-element-interface-extensions hiwch points to the `as=""` values: https://fetch.spec.whatwg.org/#concept-request-destination

...there doesn't seen to be a type for json.  IIRC, `as=""` is a hint to the browser on which parser it should use to download + parse the resource. I'd assume "script" would be appropriate for json, but not sure.  If Chrome complains, it could also be a bug in our push stack.

FredKSchott:
@ebidel thanks for the review, I didn't realize that http2-push-manifest was using hydrolysis. Would love to help upgrade to the new analyzer once Polymer 2.0 is released, potentially in a way that polymer-build could leverage/benefit from.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!--## Unreleased-->
 
 <!-- List New Changes Here -->
+* New Feature: add automatic [push manifest](https://github.com/GoogleChrome/http2-push-manifest) generation for HTTP2 push-enabled servers.
```
justinfagnani: "push manifest" -> "HTTP/2 push manifest"

README.md
```diff
@@ -160,6 +160,40 @@ addServiceWorker({
 ```
 
 
+### Generating a Push Manifest
```
justinfagnani: HTTP/2 Push Manifest

README.md
```diff
@@ -160,6 +160,40 @@ addServiceWorker({
...
+
+Polymer-Build can automatically generate a [push manifest](https://github.com/GoogleChrome/http2-push-manifest) for your application. This JSON file can be read by your http2 push-enabled web server to more easily construct the appropriate `Link: <URL>; rel=preload; as=<TYPE>` headers(s) for http2 push/preload.
```
justinfagnani: Polymer-Build -> `polymer-build`

CHANGELOG.md
justinfagnani: here and other places...

README.md
```diff
@@ -160,6 +160,40 @@ addServiceWorker({
...
+
+> **Note:** Push manifests can only be generated for app-shell projects (projects that define a shell and fragments). The manifest describes the following behavior: Requesting the shell should push any shell dependencies as well. Requesting a fragment should push any dependencies of that fragment *that were not already pushed by the shell.*
```
justinfagnani: why? for an entrypoint-only project, shouldn't we push all the dependencies for the entrypoint?

README.md
justinfagnani: capitalize http2 to HTTP/2

justinfagnani: Probably mention that the manifest is generated for the files in the build output - you'll get a different manifest per build.

README.md
```diff
@@ -160,6 +160,40 @@ addServiceWorker({
...
+
+
+#### generatePushManifest()
+
+`generatePushManifest()` will generate a push manifest for your application.
+
+```js
+const generatePushManifest = require('polymer-build').generatePushManifest;
```
justinfagnani: ultra-nit: `const {generatePushManifest} = require('polymer-build');` (assuming node > 4)

README.md
```diff
@@ -160,6 +160,40 @@ addServiceWorker({
...
+
+generatePushManifest({
+  buildRoot: 'build/',
+  project: project,
+}).then(() => { // ...
+```
+
+#### addPushManifest()
+
+Like `generatePushManifest()`, but writes the generated service worker to the file path you specify in the `path` option ("push-manifest.json" by default).
```
justinfagnani: I don't understand the difference from this description. What's generatePushManifest do if not writing a file? What are the resolved promise values for each function?

src/push-manifest.ts
```diff
@@ -0,0 +1,163 @@
...
+import {Analyzer, FSUrlLoader} from 'polymer-analyzer';
+
+import {urlFromPath} from './path-transformers';
+import {PolymerProject} from './polymer-project';
+
+
+export interface AddPushManifestOptions {
+  project: PolymerProject;
+  buildRoot: string;
+  bundled?: boolean;
```
justinfagnani: what's `bundled` for?

src/push-manifest.ts
```diff
@@ -0,0 +1,163 @@
...
+  '.json': 'script',
+  '.svg': 'image',
+  '.webp': 'image',
+  '.woff': 'font',
+  '.woff2': 'font'
+};
+
+
+export interface PushManifestEntry {
+  [id: string]: {
```
justinfagnani: Is `id` a url or path?

src/push-manifest.ts
```diff
@@ -0,0 +1,163 @@
...
+    type?: LinkType,
+    weight: 1,
+  };
+}
+export interface PushManifest { [id: string]: PushManifestEntry; }
```
justinfagnani: Why is there an id in both `PushManifestEntry` and `PushManifest`? How are they different, what are they id's of?

src/push-manifest.ts
```diff
@@ -0,0 +1,163 @@
...
+      const analyzedImports = analyzedDocument.getByKind(
+          'import', {externalPackages: true, imported: true});
+      const analyzedImportUrls = new Set();
+      function shouldIgnoreFile(url: string) {
+        return ignoreUrls && ignoreUrls.has(url);
+      }
+
+      for (const analyzedImport of analyzedImports) {
+        if (!shouldIgnoreFile(analyzedImport.url)) {
+          analyzedImportUrls.add(analyzedImport.url);
```
justinfagnani: I think you want the resolved URL, as several imports can refer to the same file with different URLs. import.document.url will be the right one. But I also think you want to get documents, not imports since they are the things that imports point to and are resolved and dedpued.

src/push-manifest.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+/**
+ * Gets a link type from a file URL. Useful for generating a PushManifestEntry
+ * and needed for the "as=" value on Link headers.
+ */
+function
+getLinkTypeFromUrl(url: string):
+    LinkType|undefined {
+      return extensionToTypeMapping[path.extname(url)] || undefined;
```
justinfagnani: this is another reason why we want to use the resolved URL. It's possible that a resolver supports URLs without file extensions - adding them for .js files for example. We want to determine the type from the resolved file, no the references to it.

src/push-manifest.ts
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * Generate a PushManifestEntry from an array of URLs.
+ */
+function
+generatePushManifestEntries(urls: string[]):
+    PushManifestEntry {
+      const newEntry: PushManifestEntry = {};
+      for (const url of urls) {
+        newEntry[url] = {type: getLinkTypeFromUrl(url), weight: 1};
```
justinfagnani: Should we lower the weight for everything except html, css and JS?

src/push-manifest.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+/**
+ * Returns a promise that resolves when a push manifest has been generated
+ * and written to the build directory. This uses generatePushManifest() to
+ * generate a push manifest, which it then writes to the file system based on
+ * the buildRoot & path (if provided) options.
+ */
+export async function
+addPushManifest(options: AddPushManifestOptions): Promise<{}> {
```
justinfagnani: `Promise<{}>` -> `Promise<void>`

justinfagnani: why do we need this method? can't we inject the manifest into the stream and let `dest` write it out?

README.md
FredKSchott: We can make assumptions about how an app-shell application behaves, we can't do the same for all other applications. In situations where the entrypoint is served from every url, you would need to support a wildcard "*" push manifest entry so that the server would push all dependencies for every possible URL, but even that would cause lots of duplicate & unnecessary pushes as the user continued to use the site.

If the app is statically hosted, we can't tell what files are pages/entrypoints and what are dependencies of those pages.

README.md
FredKSchott: 👍 but lets hold off until we actually drop node < 6 support. This is the format for the rest of the doc so no rush to make this change just yet

README.md
FredKSchott: `generatePushManifest()` returns a push manifest, `addPushManifest()` writes a push manifest to file. Clarifying...

src/push-manifest.ts
rictic: Also note that document can be undefined if we couldn't load it.

src/push-manifest.ts
FredKSchott: whoops!

src/push-manifest.ts
FredKSchott: url, clarifying...

src/push-manifest.ts
FredKSchott: One is the url for the requested resource, one is the url for the responding resource. Clarifying...

src/push-manifest.ts
```diff
@@ -0,0 +1,163 @@
...
+      }
+      return analyzedImportUrls;
+    }
+
+/**
+ * Gets a link type from a file URL. Useful for generating a PushManifestEntry
+ * and needed for the "as=" value on Link headers.
+ */
+function
+getLinkTypeFromUrl(url: string):
```
rictic: I think we want to track the file type based on how it's referenced and only use the extension as a fallback. If you script src it, it's a script. If you html import it, it's a document. If you link rel=stylesheet it, it's a style.

This is a bit more complicated, because a file could in theory be referenced in multiple different ways, and the service worker shouldn't fall over in that case. I guess this is best effort so last-import-wins should be fine when there's conflicts.

src/push-manifest.ts
```diff
@@ -0,0 +1,163 @@
...
+        }
+      }
+      return analyzedImportUrls;
+    }
+
+/**
+ * Gets a link type from a file URL. Useful for generating a PushManifestEntry
+ * and needed for the "as=" value on Link headers.
+ */
+function
```
rictic: style note: this behavior where `function` is on a different line than the function name is fixed by adding a semicolon at the end of async toplevel function declarations

src/push-manifest.ts
FredKSchott: This is actually bad naming on my part, it's relative to the root instead of relative to the file system. but obviously a URL relative to the root is an absolute url. 🤕  Clarifying...

src/push-manifest.ts
FredKSchott: Agreed, currently handled since it's an absolute URL based on the file path (see comment above)

src/push-manifest.ts
FredKSchott: See https://github.com/GoogleChrome/http2-push-manifest#whats-a-push-manifest: only weight=1 is supported at the moment. Will add a comment/TODO for when more weights are supported.

src/push-manifest.ts
FredKSchott: THANK YOU :)

README.md
FredKSchott: Holdover from perf :)

src/push-manifest.ts
FredKSchott: @rictic is that true? The TS type definitions say it's a set of documents objects, no `undefined`s allowed.

src/push-manifest.ts
FredKSchott: We could, my understanding was that streams were for transformations and methods where for file additions. This is the pattern we started with service worker, and is more performant since it doesn't suffer from the bottlenecking problems we have with stream loading.

Happy to revisit this if there's a good reason to turn it into a stream.

README.md
FredKSchott: Clarified in `generatePushManifest()` description & code example

src/push-manifest.ts
justinfagnani: The only reason we read and write from the file system for service workers in that that's all `sw-precache` supports. `sw-lib` will allow us to just use the stream. We should just use the stream here so if someone continues to process it or redirect to another location it just works.

README.md
justinfagnani: The most common app style before we launched the CLI, and possibly still, was a single, non-routed or fragment-routed, page. We should still support that pattern, so I'd suggest that without a shell we assume that the entrypoint is only loaded from a single URL and we can build the push manifest for it just like we do the shell. In other places we treat the absence of fragments and shell as an indicator to treat the entrypoint like the shell, and this would go along with that.

src/push-manifest.ts
justinfagnani: good to know

README.md
FredKSchott: So in that example, all dependencies are pushed when `/index.html` is requested, but then the service worker caches `index.html` and intercepts future requests to the server that would repush the entire app again.

Okay, if the user is asking for a push manifest in their non-app-shell application that makes sense. +1 will add.

This also reminds me that I should add a warning if `--add-push-manifest` is included but `--add-service-worker` is not, since right now the entire pushing strategy relies on the server worker intercepting unnecessary requests.

src/push-manifest.ts
FredKSchott: but it wouldn't just work, in some cases. It needs to be run on the final application. If any files are changed after it's generated the push manifest can silently break.

src/push-manifest.ts
rictic: I meant `import.document` but actually looking at it, it seems like we might not produce an Import unless we can resolve the document? That's a bit surprising, but kinda reasonable too.

src/push-manifest.ts
```diff
@@ -0,0 +1,160 @@
...
+        }
+      }
+      return analyzedImportUrls;
+    };
+
+/**
+ * Gets a link type from a file URL. Useful for generating a PushManifestEntry
+ * and needed for the "as=" value on Link headers.
+ */
+function getLinkTypeFromUrl(url: string): LinkType|undefined {
```
rictic: I think we want to track the file type based on how it's referenced and only use the extension as a fallback. If you script src it, it's a script. If you html import it, it's a document. If you link rel=stylesheet it, it's a style.

This is a bit more complicated, because a file could in theory be referenced in multiple different ways, and the service worker shouldn't fall over in that case. I guess this is best effort so last-import-wins should be fine when there's conflicts.

FredKSchott: Lets do this then, we have all the data so it shouldn't be too much extra work. Thanks for thinking this extra bit through

src/push-manifest.ts
FredKSchott: makes sense as long as there's some other way to detect unresolved imports (I remembered them to be warnings?)

src/push-manifest.ts
```diff
@@ -41,72 +81,106 @@ const extensionToTypeMapping: {[extname: string]: LinkType} = {
...
+  return extensionToTypeMapping[path.extname(url)] || undefined;
+}
+
+/**
+ * Get the resource type for an import, handling special import types and
+ * falling back to getResourceTypeFromUrl() if none exist.
+ */
+function getResourceTypeFromImport(importFeature: Import): ResourceType|
+    undefined {
+  const importTagName = importFeature.astNode.tagName;
```
rictic: recommend switching off of checking various importFeature.kinds.has() options instead, as the astNode may not generally be an HTML node.

Hm, arguably the kind of imported resource should probably be something that the Import tells you, rather than you try to deduce.

src/push-manifest.ts
FredKSchott: Talked in person, I'm convinced. As an added bonus, the interface is now much easier to use (doesn't need a `buildRoot` because the file stream is all relative).

src/push-manifest.ts
FredKSchott: I was about to file an issue about that in the analyzer :) The type of language/astNode format is something that I'd like to check before accessing it. Checking the existence of the `tagName` property was the quick and dirty solution I came up with, but I like checking `kinds` much better.

FredKSchott: Arg nevermind. `kinds` describes the import and not the importer, so it doesn't help you figure out what kind of astNode you're dealing with (ex: `<link rel="stylesheet" ...` has kinds `'import', 'html-style'`). I'll keep my workaround for now with a NOTE and then follow up in an analyzer issue for a potential enhancement

FredKSchott:
@justinfagnani @rictic PTAL
- new stream interface
- support bundled applications
- support non-app-shell applications
- updated readme

src/push-manifest.ts
rictic: I think you shouldn't have to check the ast node at all. `html-style` and `css-import` tells you that the file is css, `html-import` tells you that it's an html document, `html-script` tells you that it's a javascript file. `js-import` is a bit more interesting, amusingly, and there we should probably check the filename because an ES6 import's semantics hinge on the mime type it's served with. The others though, I think believe that they're blindly trusted to be the expected content type, so we can use the same semantics.

FredKSchott: @rictic interesting, can you share more info about ES6 imports & mime type? I had understood that `import` could only import JS modules, unless you're referring to babel craziness.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!--## Unreleased-->
 
 <!-- List New Changes Here -->
+* New Feature: add automatic [push manifest](https://github.com/GoogleChrome/http2-push-manifest) generation for HTTP/2 Push-enabled servers.
```
justinfagnani: "push manifest" -> "HTTP/2 push manifest"

README.md
```diff
@@ -160,6 +160,29 @@ addServiceWorker({
 ```
 
 
+### Generating an HTTP/2 Push Manifest
+
+`polymer-build` can automatically generate a [push manifest](https://github.com/GoogleChrome/http2-push-manifest) for your application. This JSON file can be read by your HTTP/2 push-enabled web server to more easily construct the appropriate `Link: <URL>; rel=preload; as=<TYPE>` headers(s) for HTTP/2 push/preload.
```
justinfagnani: Might be good to drop in a link to one of the sample servers that support this. It'd be my first question.

README.md
```diff
@@ -105,7 +105,7 @@ You can add splitters to any part of your build stream. We recommend using them
 
 #### project.bundler()
 
-A stream that combines the files in your application to reduce the number of frontend requests needed. This can be a great way to [improve performance](https://developer.yahoo.com/performance/rules.html#num_http) when HTTP2/Push is not available.
+A stream that combines the files in your application to reduce the number of frontend requests needed. This can be a great way to [improve performance](https://developer.yahoo.com/performance/rules.html#num_http) when HTTP/2 Push is not available.
```
justinfagnani: I'd remove the "when..." part, since our own experiments show that bundling is always a big win.

README.md
```diff
@@ -160,6 +160,29 @@ addServiceWorker({
...
+### Generating an HTTP/2 Push Manifest
+
+`polymer-build` can automatically generate a [push manifest](https://github.com/GoogleChrome/http2-push-manifest) for your application. This JSON file can be read by your HTTP/2 push-enabled web server to more easily construct the appropriate `Link: <URL>; rel=preload; as=<TYPE>` headers(s) for HTTP/2 push/preload.
+
+The generated push manifest describes the following behavior: Requesting the shell should push any shell dependencies as well. Requesting a fragment should push any dependencies of that fragment *that were not already pushed by the shell.* If no shell was defined for your build, `polymer-build` will use the application entrypoint URL instead (default: `index.html`).
+
+
+#### project.addPushManifest()
+
+This method will return a stream that injects a new push manifest into your build (default: `push-manifest.json`). The push manifest is based off the application import graph, so make sure that this stream is added after all changes to the application source code.
```
justinfagnani: "stream" -> "transform"

src/push-manifest.ts
```diff
@@ -0,0 +1,213 @@
...
+export interface PushManifestEntry {
+  type?: ResourceType;
+  weight: 1;
+}
+export interface PushManifestEntryCollection {
+  [dependencyAbsoluteUrl: string]: PushManifestEntry;
+}
+export interface PushManifest {
+  [requestAbsoluteUrl: string]: PushManifestEntryCollection;
+}
```
justinfagnani: Big 👍 to the name changes. Very clear now.

src/push-manifest.ts
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * A mapping of file extensions and their default resource type.
+ */
+const extensionToTypeMapping: {[extname: string]: ResourceType} = {
```
justinfagnani: I'd use a `Map` - it's always safer.

src/push-manifest.ts
```diff
@@ -0,0 +1,213 @@
...
+  '.webp': 'image',
+  '.woff': 'font',
+  '.woff2': 'font'
+};
+
+/**
+ * Get the default resource type for a file based on its extension.
+ */
+function getResourceTypeFromUrl(url: string): ResourceType|undefined {
+  return extensionToTypeMapping[path.extname(url)] || undefined;
```
justinfagnani: `|| undefined` shouldn't be necessary

src/push-manifest.ts
```diff
@@ -0,0 +1,213 @@
...
+ * Analyze the given URL and resolve with a collection of push manifest entries
+ * to be added to the overall push manifest.
+ */
+async function generatePushManifestEntryForUrl(
+    analyzer: Analyzer, url: string, ignoreUrls?: string[]):
+    Promise<PushManifestEntryCollection> {
+      const analyzedDocument = await analyzer.analyze(url);
+      const analyzedImports = analyzedDocument.getByKind(
+          'import', {externalPackages: true, imported: true});
+      const analyzedImportUrls: PushManifestEntryCollection = {};
```
justinfagnani: use a `Map`

README.md
FredKSchott: +1 great idea

README.md
FredKSchott: Not everyone knows that "transform" means stream, but "stream" -> "transform stream" could be helpful for the first mention. Making that change

src/push-manifest.ts
FredKSchott: Since we're constructing a JSON object I'd recommend keeping this as a JS object. Converting map-to-JS-object after the fact requires an extra step/iteration over the map and  the practical value added by a Map in this case is minimal.

FredKSchott:
@justinfagnani re tests: yes! Was waiting until the interface had been through review, adding tests now

FredKSchott:
@justinfagnani PTAL
- responded to formatting review comments
- added tests

src/push-manifest.ts
justinfagnani: ah, might be clearer if this was named `pushManifestEntries` then

src/push-manifest.ts
```diff
@@ -0,0 +1,213 @@
...
+
+      return pushManifestEntries;
+    };
+
+
+/**
+ * A stream that reads in files from an application to generate an HTTP2/Push
+ * manifest that gets injected into the stream.
+ */
+export class AddPushManifest extends Transform {
```
slightlyoff: It's really weird to have a verb in the name of a class

PR #177, opened 2017-04-13
Move the native-CE + ES5 shim transformer from cli to build.

aomarks:
Renamed to inject-es5-adapter.ts so that the file matches the class.

Original: https://github.com/Polymer/polymer-cli/blob/master/src/build/rewrite-webcomponents-loader.ts


aomarks:
Abandoning this PR for now, going to expose the interface differently.

PR #180, opened 2017-04-13
Update Polymer Analyzer

maxknee:


rictic:
We'll be working on this very soon, but the analyzer update is a major breaking change, so we'll need to update the places we use it in build.

usergenic:
FWIW, this is happening here https://github.com/Polymer/polymer-build/pull/184

PR #181, opened 2017-04-13
Expose Custom Elements ES5 Adapter from polymer-build.

aomarks:
- NOTE: The 1st commit here copies custom-elements-es5-adapter.js wholesale from cli to build, and the 2nd has all the changes to it. So the 2nd commit is the more interesting diff to look at.
- Expose as a stream transformer via a method on PolymerProject. To be used by polymer-cli.
- Also expose as a pure HTML string function. To be used by polyserve. Documented to dissuade users from using it.
- Part 1/3 of https://github.com/Polymer/polyserve/issues/164. After this is released, we can update polymer-cli and polyserve.


justinfagnani:
Thanks for organizing the commits!

src/custom-elements-es5-adapter.ts
```diff
@@ -31,22 +31,10 @@ const styleMatcher = p.AND(
...
- * NOTE(fks) 03-28-2017: Comments are not currently moved to the body, which
- * means relevent comments may be left behind in the head. This is fine for now,
- * but may break if we ever decide to support important metadata in comments
- * (ex: support for defining lazy-imports via HTML comments).
- *
- * TODO(fks) 03-28-2017: Move into polymer-build and add tests.
+ * Wraps `addCustomElementsEs5Adapter()` in a `stream.Transform`.
  */
-export class InjectES5Adapter extends stream.Transform {
+export class CustomElementsEs5AdapterTransformer extends stream.Transform {
```
justinfagnani: such a great bikeshedding opportunity: nice noun phrase :) but we don't need to append the type, we can just use the description of what it does, like `CustomElementsEs5AdapterInjector`

aomarks: Done.

Issue #165, opened 2017-03-21
[Feature Request] Allow to configure minEntrypoints passed to generateShellMergeStrategy


web-padawan:
As @usergenic said ealirer [here](https://github.com/Polymer/polymer-build/pull/108#discussion_r96712278), adding configuration option for [this line](https://github.com/Polymer/polymer-build/blob/master/src/bundle.ts#L78) should be separate PR, but is probably worth looking at soon.

Do you have any plans about this? Is it a future work after releasing `polymer-build` 1.0.0 ?

I'm interested in this stuff because of my experience with a large project. There are lots of dependencies only required by 2 fragments, but as of now, they are all being bundled into the shell. Thanks to libraries like `ace` and `moment` (without locales), our shell JS bundle is more than 300 K minified and gzipped, even though we try to lazy load as much as we can.

Related to https://github.com/Polymer/polymer-bundler/issues/437

justinfagnani:
@usergenic we'll need this added to the project config and CLI too.

usergenic:
@justinfagnani I'm not sure it makes sense to add the minEntrypoints directly to the config.  I think it makes more sense to have the user provide the argument like this in their gulp-file:
```js
.pipe(project.bundle({strategy: generateShellMergeStrategy('shell.html', 3)}))
```

That would work with https://github.com/Polymer/polymer-build/pull/189

usergenic:
polymer-build 1.2.0 published 🎉

PR #189, opened 2017-04-18
Added options to `polymer-project`'s `bundler()` method.

usergenic:
 - Added an options argument to `PolymerProject#bundler()` to support users
configuring the bundler.  See `Options` in `src/build-bundler`.
 - Calculate the files that change since in the Analyzer has seen them so we can `_fork` the Analyzer now instead of creating all new one.
  - Fixed 2 bugs in the test code around setting up piped streams and transforming files.


web-padawan:
Hey @usergenic! Would you mind adding config for `minEntrypoints` as described in #165 as well? Or it would be better to send a separate PR for that?

jjuutila:
Could you add some documentation about the options to the readme too?

usergenic:
@jjuutila wrt documentation: yes.  will get that in asap.

I've got a fairly big update to do for `bundler`'s README to describe how to use `BundleStrategy`s and `BundleUrlMapper`s.  All of the options are options for `Bundler` constructor.

@web-padowan with this interface you can now provide a bundle strategy directly with whatever minEntrypoints you want:

```javascript
var myShellStrategy = require('polymer-bundler/lib/bundle-manifest').generateShellMergeStrategy('shell.html', 3);

...
.pipe(project.bundler({strategy: myShellStrategy}));
...
```

usergenic:
@jjuutila just updated the `bundle_test.ts` with several illustrative cases for each option.

src/bundle.ts
```diff
@@ -14,45 +14,75 @@
...
   files = new Map<string, File>();
 
-  constructor(config: ProjectConfig, buildAnalyzer: BuildAnalyzer) {
+  constructor(
+      config: ProjectConfig,
+      buildAnalyzer: BuildAnalyzer,
+      options?: Options) {
     super({objectMode: true});
 
+    if (!options) {
```
justinfagnani: Use a default parameter

src/bundle.ts
```diff
@@ -14,45 +14,75 @@
...
+         strategy,
+         urlMapper} = options;
+
+    const urlLoader = new FileMapUrlLoader(
+        this.config.root, this.files, analyzer || buildAnalyzer.analyzer);
+
+    analyzer = analyzer ? analyzer._fork({urlLoader}) :
+                          buildAnalyzer.analyzer._fork({urlLoader});
+
+    inlineCss = typeof inlineCss === 'undefined' || inlineCss;
```
justinfagnani: Just `inlineCss === undefined`. If someone ever defines `undefined` in a module they should get a peer unbonus

justinfagnani: Also, destructuring assignment supports defaults: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

so just do:

```
let {
  inlineCss = true,
  inlineScripts = true,
} = options;
```

That should be much clearer.

src/bundle.ts
```diff
@@ -14,45 +14,75 @@
...
+         inlineCss,
+         inlineScripts,
+         rewriteUrlsInTemplates,
+         sourcemaps,
+         stripComments,
+         strategy,
+         urlMapper} = options;
+
+    const urlLoader = new FileMapUrlLoader(
+        this.config.root, this.files, analyzer || buildAnalyzer.analyzer);
```
justinfagnani: There seems to be a circular dependency here between urlLoader and analyzer. Why's that? It's a little confusing to follow how has references to who between this and the line below. If this is needed, maybe use some new variables names and don't reassign to `analyzer`?

src/bundle.ts
```diff
@@ -83,14 +122,16 @@ export class BuildBundler extends Transform {
   }
 
   async _buildBundles(): Promise<Map<string, string>> {
-    let strategy: BundleStrategy;
-    if (this.config.shell) {
+    let strategy = this._strategy;
```
justinfagnani: consider making `strategy` const and using a ternary

src/file-map-url-loader.ts
```diff
@@ -23,21 +23,31 @@ import {parseUrl} from 'polymer-analyzer/lib/utils';
 export class FileMapUrlLoader implements UrlLoader {
```
justinfagnani: Don't we have one of these now in the Analyzer? TODO to use that?

src/test/bundle_test.ts
```diff
@@ -35,7 +37,7 @@ class FileTransform extends Transform {
     this.transform = transform;
   }
   _transform(f: File, _encoding: string, cb: () => void) {
-    this.transform(this, f);
+    this.transform(this, f.clone());
```
justinfagnani: was this fixing a separate bug?

src/bundle.ts
usergenic: Done.

src/bundle.ts
usergenic: I'm doing essentially the identical thing that the `InMemoryOverlayUrlLoader` does in Analyzer; here I just have need to use `File` objects as opposed to content strings which is why I'm not using the former.  Both loaders require a fallback, and since I can't fallback to the previous loader, because it is not exposed, I have to fallback to the previous Analyzer, which implements it.

However, you raise a fair point about the variable naming in that reassigning analyzer may be obfuscating what's happening.  I've adjusted that.

src/bundle.ts
usergenic: Defaults look totes so much better.  I'll remember that.  Fixed.

src/bundle.ts
usergenic:
```typescript
const strategy = this._strategy ||
        this.config.shell && generateShellMergeStrategy(urlFromPath(
                                 this.config.root, this.config.shell));
```

Went with the above over ternary because even shorter.

src/file-map-url-loader.ts
usergenic: The Analyzer's one is similar but I preferred to use this loader because I could preserve the File objects and not have to `toString()` all of their buffer contents as I loaded the map.  A subsequent PR could look to unify behind `InMemoryOverlayUrlLoader` but I didn't think it bought much to overhaul it.  I did actually start down the path of doing so, but aborted.

src/test/bundle_test.ts
usergenic: This just addressed a bug in the tests themselves.  The old tests passed either way, but they didn't prove what they needed to without the clone.

usergenic:
@justinfagnani feedback addressed PTAL

src/bundle.ts
```diff
@@ -14,45 +14,70 @@
...
+         strategy,
+         urlMapper} = options;
+
+    const urlLoader = new FileMapUrlLoader(
+        this.config.root, this.files, analyzer || buildAnalyzer.analyzer);
+
+    const forkedAnalyzer = analyzer ? analyzer._fork({urlLoader}) :
+                                      buildAnalyzer.analyzer._fork({urlLoader});
+
+    this._bundler = new Bundler({
```
FredKSchott: reconstructing the options manually with a whitelist adds a lot of maintenance overhead to track changes to BundleOptions (we saw this in serve & test on the CLI). Instead pass a new options object based on the one given, like:

```js
const bundlerOptions = Object.assign({}, options, {
  analyzer: forkedAnalyzer,
  inlineCss: true,
  inlineScripts: true,
  strategy: undefined,
  urlMapper: undefined,
});
this._bundler = new Bundler(bundlerOptions);
```

src/file-map-url-loader.ts
```diff
@@ -23,21 +23,31 @@ import {parseUrl} from 'polymer-analyzer/lib/utils';
...
+  // Try to load the file from the map.  If not in the map, try to load
+  // from the fallback loader.
   async load(url: string): Promise<string> {
     const file = this.files.get(parseUrl(url).pathname)!;
 
     if (file == null) {
+      if (this.fallbackLoader && this.fallbackLoader.canLoad(url)) {
+        return this.fallbackLoader.load(url);
+      }
       throw new Error(`File ${url} not present in file map.`);
```
FredKSchott: update error below to include info if fallback exists but `!canLoad`

src/polymer-project.ts
```diff
@@ -96,8 +97,10 @@ export class PolymerProject {
   }
 
   /**
-   * Returns a stream transformer that injects `custom-elements-es5-adapter.js`
-   * into the entry point HTML file. This adapter is needed when serving ES5 to
+   * Returns a stream transformer that injects
```
FredKSchott: bad clang?

src/test/bundle_test.ts
```diff
@@ -391,4 +394,108 @@ suite('BuildBundler', () => {
...
+  suite('options', () => {
+
+    const projectOptions = {
+      root: 'test-fixtures/test-project',
+      entrypoint: 'index.html',
+      fragments: ['shell.html'],
+      sources: ['index.html', 'shell.html', 'source-dir/my-app.html']
+    };
+
+    test('excludes: a file', async() => {
```
FredKSchott: we haven't decided on this as a team so totally up to you, but I prefer test description to explain the intention / what's being tested. 

```diff
-test('excludes: a file', async() => {
+test('excludes a file from bundling when options.excludes includes that file path', async() => {
```

(to prove this point a bit, I actually had to re-read the contents of the test to understand what was being tested, since I incorrectly assumed an excluded file wouldn't be returned by `getFile()`)

src/test/polymer-project_test.ts
```diff
@@ -78,10 +80,18 @@ suite('PolymerProject', () => {
     assert.isTrue(getFlowingState(dependencyStream));
   });
 
-  test('the bundler method returns a different bundler each time', () => {
-    const bundlerA = defaultProject.bundler();
-    const bundlerB = defaultProject.bundler();
-    assert.notEqual(bundlerA, bundlerB);
+  suite('.bundler()', () => {
+
+    test('returns a different bundler each time', () => {
```
FredKSchott: smart test!

src/test/polymer-project_test.ts
```diff
@@ -78,10 +80,18 @@ suite('PolymerProject', () => {
...
+      const bundlerA = defaultProject.bundler();
+      const bundlerB = defaultProject.bundler();
+      assert.notEqual(bundlerA, bundlerB);
+    });
+
+    test('takes options to configure bundler', () => {
```
FredKSchott: incorrect description? are you testing that this can take an empty object without failing?

src/bundle.ts
justinfagnani: No need to change, but... :)

Ternaries are a little better because they avoid boolean conversion, and sometimes the logic is easier to follow than compound logical expressions:

```ts
const strategy = this._strategy !== undefined
    ? this._trategy
    : this.config.shell !== undefined
        ? generateShellMergeStrategy(urlFromPath(this.config.root, this.config.shell));
        : undefined;
```

I can read this as if/else/if/else...

src/file-map-url-loader.ts
```diff
@@ -23,21 +23,31 @@ import {parseUrl} from 'polymer-analyzer/lib/utils';
...
+      files: Map<string, File>,
+      fallbackLoader?: UrlLoader) {
     this.root = root;
     this.files = files;
+    this.fallbackLoader = fallbackLoader;
   }
 
   // We can always return true because we're just reading paths off a map.
   canLoad(_url: string): boolean {
     return true;
```
justinfagnani: shouldn't this return either `this._fallbackLoader.canLoad(url)`, `this.files.has(parseUrl(url).pathname) || this._fallbackLoader.canLoad(url)`, or something related to `root`?

src/file-map-url-loader.ts
```diff
@@ -23,21 +23,31 @@ import {parseUrl} from 'polymer-analyzer/lib/utils';
 export class FileMapUrlLoader implements UrlLoader {
   root: string;
   files: Map<string, File>;
+  fallbackLoader?: UrlLoader;
 
-  constructor(root: string, files: Map<string, File>) {
+  constructor(
+      root: string,
```
justinfagnani: `root` appears to never be used

src/file-map-url-loader.ts
```diff
@@ -23,21 +23,31 @@ import {parseUrl} from 'polymer-analyzer/lib/utils';
 export class FileMapUrlLoader implements UrlLoader {
   root: string;
   files: Map<string, File>;
+  fallbackLoader?: UrlLoader;
```
justinfagnani: can these be private? You might get unused errors if they are.

src/test/bundle_test.ts
justinfagnani: This is a meaningful change in behavior though, especially with the potential forest of streams we'll encounter. Can a test fail without this?

src/test/bundle_test.ts
justinfagnani: I generally prefer that 1) the nested suite names + test name read mostly like a sentence that describes the intention of the test 2) The suite names mostly refer to concrete elements in the source - libraries, functions, classes, methods, etc. 3) lower priority, avoid too much prose to make the combined description hard to parse, use terse, list-item like wording.

In this case the tests are running through each option, so for readability I like having them all start with the option name. It's easier to parse the whole suite that way.

src/bundle.ts
justinfagnani: I think the thing that's throwing me here is that `buildAnalyzer.analyzer` would be passed to a `UrlLoader`. That's seems inverted from a scope of responsibilities POV. I guess `Analyzer` implements `UrlLoader`? Why doesn't it just have one?

src/bundle.ts
FredKSchott: That's funny, I see the type checker as exactly why this doesn't carry the same risk as it might in pure JS. Type checker prevents incorrect options from being passed, and updates  automatically with the BundlerOptions type. I don't see any reason to be wary.

src/bundle.ts
usergenic: @justinfagnani are you why Analyzer doesn't have a read-only `urlLoader` public property?  Analyzer actually implements both `UrlLoader` and `UrlResolver` now I think.  It's a fair question why it might not want to simply expose the loader and resolver instead.  When it exposes loader, I'll just use that here.

src/bundle.ts
usergenic: Well, there's a couple of things re: the options.  Unlike CLI, Build doesn't really just act as a agnostic pass-thru for this stuff.  For example, every option here has 2 tests in the Build's test suite.  I think for now, I'd like to keep this explicit, though I do get what you're saying.

src/bundle.ts
```diff
@@ -14,45 +14,70 @@
...
-      // to mean, at least, a doubling of analysis efforts for bundling phase.
-      // Ideally we would fork existing analyzer and replace its urlLoader if
-      // there were an affordance to do so.
-      analyzer: new Analyzer(
-          {urlLoader: new FileMapUrlLoader(this.config.root, this.files)}),
-      inlineCss: true,
-      inlineScripts: true,
+    let {analyzer,
+         excludes,
+         inlineCss = true,
```
usergenic: Note: This inlineCss = true won't be necessary after next bundler release since inlining will be true by default.

src/test/bundle_test.ts
usergenic: Here is how the options report in context for the `BuildBundler` suite

```
BuildBundler
    ✓ entrypoint only
    ✓ two fragments (38ms)
    - shell and entrypoint
    ✓ shell and fragments with shared dependency (48ms)
    - entrypoint and fragments
    ✓ bundler loads changed files from stream (63ms)
    ✓ bundler deals with win32 platform separators on win32 (47ms)
    ✓ bundler deals with posix platform separators on win32
    ✓ bundler does not output imports which have been bundled
    options
      ✓ excludes: a file (49ms)
      ✓ excludes: nothing
      ✓ inlineCss: false (38ms)
      ✓ inlineCss: true (43ms)
      ✓ inlineScripts: false
      ✓ inlineScripts: true (40ms)
      ✓ stripComments: false (49ms)
      ✓ stripComments: true (44ms)
      ✓ strategy: fn() (45ms)
      ✓ urlMapper: fn() (49ms)
```

usergenic: I'll update the test names to include description of expected behavior. @FredKSchott @justinfagnani:

```
options
      ✓ excludes: html file urls listed are not inlined (51ms)
      - excludes: html files in folders listed are not inlined
      ✓ excludes: nothing is excluded when no excludes are given (40ms)
      ✓ inlineCss: false, does not inline external stylesheets
      ✓ inlineCss: true, inlines external stylesheets (41ms)
      ✓ inlineScripts: false, does not inline external scripts
      ✓ inlineScripts: true, inlines external scripts
      ✓ stripComments: false, does not strip html comments
      ✓ stripComments: true, strips html comments
      ✓ strategy: fn(), applies bundle strategy function
      ✓ urlMapper: fn(), applies bundle url mapper function
```

src/test/bundle_test.ts
usergenic: The main reason this clone was necessary is that tests using this transform were trying to demonstrate that the BuildBundler was processing value from the incoming stream, but without cloning the file in the transform here in the test, the test would pass if loader read the old file since the contents were changed in the file.  By forcing the clone, I demonstrate that the BuildBundler is processing the file that's coming in on the stream.  There's nothing _AFAIK_ that would break if the `clone()` was removed.  Additionally, it should be established best practice for transform streams to produce new objects instead of mutating existing ones if they are going to support parallelization etc, so I don't think this means the underlying behavior expected is any different.

src/file-map-url-loader.ts
usergenic: Done.

src/polymer-project.ts
usergenic: Fixed.  Thanks!

src/test/polymer-project_test.ts
usergenic: 👍

src/test/polymer-project_test.ts
usergenic: I guess this description was a hold-over from when it took a bunch of options and they were defined.  I've updated it to show it takes the actual options now.

src/file-map-url-loader.ts
usergenic: Removed.

src/bundle.ts
```diff
@@ -68,6 +93,15 @@ export class BuildBundler extends Transform {
   }
 
   async _flush(done: (error?: any) => void) {
+    const filesChanged: string[] = [];
```
usergenic: I moved some code out into methods for this to clarify intent.

usergenic:
@justinfagnani Addressed all PR feedback.  PTAL!

src/file-map-url-loader.ts
usergenic: Okay I updated to use the map and the fallbackLoader (if present) in determining canLoad.

usergenic:
This branch now includes latest bundler and analyzer as well.

stramel:
:tada:

PR #196, opened 2017-05-03
Upgrade bundler to 2.0.0 pre.14

usergenic:
- Dependency update for polymer-bundler.
 - Made some methods on BuildBundler explicitly private that were written as private.
 - Removed private `_strategy` and `_urlMapper` properties on BuildBundler that were redundant now that Bundler API changed and no longer required them to be squirreled away for subsequent call to generateManifest.

### DIFF
--- a/src/bundle.ts
+++ b/src/bundle.ts
@@ -14,8 +14,7 @@
 
 import File = require('vinyl');
 import * as parse5 from 'parse5';
-import {Bundler, Options as BundlerOptions} from 'polymer-bundler';
-import {BundleManifest, BundleStrategy, generateShellMergeStrategy, BundleUrlMapper} from 'polymer-bundler/lib/bundle-manifest';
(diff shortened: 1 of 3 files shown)


src/bundle.ts
```diff
@@ -67,20 +62,24 @@ export class BuildBundler extends Transform {
     const forkedAnalyzer = analyzer ? analyzer._fork({urlLoader}) :
                                       buildAnalyzer.analyzer._fork({urlLoader});
 
+    strategy = strategy ||
```
rictic: This default moved up from bundler?

usergenic: The strategy is now a constructor argument for the Bundler.  It was previously handed to generateManifest.  Was that your question?

rictic: Oh, I didn't notice that this was just moving this code here from `_generateBundleManifest`. LGTM

PR #198, opened 2017-05-03
Minor fixup to the sample code for bundler in README.

usergenic:
```js
const {generateCountingSharedBundleUrlMapper,
       generateSharedDepsMergeStrategy} = require('polymer-bundler');

mergeStream(project.sources(), project.dependencies())
  .pipe(project.bundler({
    excludes: ['bower_components/polymer-code-mirror'],
    sourcemaps: true,
    stripComments: true,
    strategy: generateSharedDepsMergeStrategy(3),
    urlMapper: generateCountingSharedBundleUrlMapper('shared/bundle_')
  }))
  .pipe(gulp.dest('build/'));
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,13 +122,18 @@ mergeStream(project.sources(), project.dependencies())
 The bundler() method accepts an options object to configure bundling.  See [Using polymer-bundler programmatically](https://github.com/polymer/polymer-bundler#using-polymer-bundler-programmatically) for a detailed list of accepted options.
 
 ```js
+const {generateCountingSharedBundleUrlMapper,
+       generateSharedDepsMergeStrategy} = require('polymer-bundler');
+
+mergeStream(project.sources(), project.dependencies())
   .pipe(project.bundler({
     excludes: ['bower_components/polymer-code-mirror'],
     sourcemaps: true,
     stripComments: true,
-    strategy: require('polymer-bundler/lib/bundle-manifest')
-      .generateSharedDepsMergeStrategy(3)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -122,13 +122,18 @@ mergeStream(project.sources(), project.dependencies())
...
+
+mergeStream(project.sources(), project.dependencies())
   .pipe(project.bundler({
     excludes: ['bower_components/polymer-code-mirror'],
     sourcemaps: true,
     stripComments: true,
-    strategy: require('polymer-bundler/lib/bundle-manifest')
-      .generateSharedDepsMergeStrategy(3)
+    strategy: generateSharedDepsMergeStrategy(3),
+    urlMapper: generateCountingSharedBundleUrlMapper('shared/bundle_')
```
FredKSchott: is there a default behavior in polymer-bundler if none is provided? This is getting a bit into how the sausage is made, which is fine if its required but adds confusion unnecessarily if this isn't necessary for most people.

usergenic: I think its ultimately good to show that bundler's operations can be tweaked, so while people don't typically need to change strategies, its good to see how its done, since its not all that obvious.

The default behavior should be implicit since we show an example without options.  It so happens that if no strategy is specified and a shell is in the config, build will tell bundler to use a shell merge strategy, otherwise there is no additional strategy applied.

I want to show urlMapper since it allows users to name and relocate generated bundles, since "./shared_bundle_1.html" is probably irksome to people for any number of reasons.

FredKSchott: 👍  thanks for the thoughtful answer!

PR #206, opened 2017-05-11
prepare v1.2.4

FredKSchott:
Gets the es5 shim out to the people

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ mergeStream(project.sources(), project.dependencies())
 
 ### Custom Elements ES5 Adapter
 
-If your build pipeline outputs ES5 custom elements (either from source or by compilation with a tool like Babel), you must include the [Custom Elements ES5 Adapter](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js). The ES5 adapter allows compiled elements work on browsers that expect ES2015 native custom elements. See the [adapter documentation](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js) for details of why this is necessary.
+If your build pipeline outputs ES5 custom elements (either from source or by compilation with a tool like Babel), you must include the [Custom Elements ES5 Adapter](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js). The ES5 adapter allows compiled elements to work on browsers that expect ES2015 native custom elements. See the [adapter documentation](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js) for details of why this is necessary.
 
 #### project.addCustomElementsEs5Adapter()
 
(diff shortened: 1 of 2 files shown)


stramel:
Will this break people using other HTML minifiers in a gulp process?

justinfagnani:
@stramel it shouldn't, `<!--! -->` comments should be preserved by minifiers. We hope.

FredKSchott:
Yea, it's a rare use case:

- using an html minifier...
- that doesn't support the `<!--!` format for comments that should be preserved (all but 1 that I tested did not)...
- AND runs after the `addCustomElementsEs5Adapter()` stream

But as rare as it is I agree that by strict semver scripture, that's a major change. And version numbers are free. Hmm... 

At the very least, I'm going to document this better in the CHANGELOG and add a note in the README while I think about this some more.

stramel:
Yeah it appears that at least the top HTML minifiers for gulp all use the kangax minifier as their core. Only one seemed home grown and hasn't been touched in about a year from what I saw. 

It does seem rare that it could happen. Perhaps, just documenting that the minifier has to support ignoring `<!--!`? /shrug it was just a thought

FredKSchott:
Keeping this as an internal change / bug fix / patch so that people get the fix on Chrome automatically. I'm confident that no one is placing this stream before their minifiers. And on the off chance that I'm wrong and someone is, I'm equally confident that they are using `html-minifier` or another of the major minifiers that supports `<!--!`. 

Will include a note in #tools channel announcement as well.

FredKSchott:
@justinfagnani updated docs, ptal?

README.md
```diff
@@ -200,11 +200,13 @@ mergeStream(project.sources(), project.dependencies())
 
 ### Custom Elements ES5 Adapter
 
-If your build pipeline outputs ES5 custom elements (either from source or by compilation with a tool like Babel), it is critical to include the [Custom Elements ES5 Adapter](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js). This adapter provides compatibility between custom elements defined as ES5-style classes and browsers with native implementations of the Custom Elements API, such as Chrome. See the [adapter documentation](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js) for details of why this is necessary.
+If your build pipeline outputs ES5 custom elements (either from source or by compilation with a tool like Babel), you must include the [Custom Elements ES5 Adapter](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js). The ES5 adapter allows compiled elements work on browsers that expect ES2015 native custom elements. See the [adapter documentation](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js) for details of why this is necessary.
```
stramel: The ES5 adapter allows compiled elements _to_ work on browsers that expect ES2015 native custom elements.

stramel:
Other than that LGTM :+1:

README.md
FredKSchott: whoops, it's been a long day. Good catch :)

PR #207, opened 2017-05-11
Add an entrypoint base tag updater.

aomarks:
One of the features needed for differential serving. The app structure this has in mind is described here: https://github.com/Polymer/shop/pull/121

See this README for info and https://github.com/Polymer/polymer-project-config/pull/21 for how it might be integrated with `polymer build`.


usergenic:
@aomarks What files get piped through this?  We should consider the possibility of files which contain their `<base>` tags *with* hrefs which should be resolved appropriately instead of just setting it verbatim.

For example, consider an html file at: `compiled-and-bundled/src/cart/cart.html` has a relative `<base href="../">` and the rewrite base url is `/compiled-and-bundled/` then it should resolve correctly to `/compiled-and-bundled/src/`.

Existing absolute base hrefs are harder to decide, but given the builds are intended to be served on their own base mounts, I'd treat absolute to the base href, so `<base href="/src/">` should be written as `/compiled-and-bundled/src/`.

Does that make sense?

usergenic:
I recognize this is intended purely for entrypoint files, but if it doesn't address base tags on fragments then any app that has them will have issues.

aomarks:
Good point. At least with the app structure this PR has in mind, though, I _think_ it works out so fragments don't need transforming. (BTW I just updated this PR description with a link to a shop PR which hopefully gives a little more context.)

The reason fragments should be OK is that they are served from fully qualified paths (e.g. at `/bundled-es5/fragment.html`), hence their relative hrefs already resolve correctly, including with <base> tags.

So, it should be only the entrypoint that needs massaging, because it's served from `/` and `/made/up/application/routes`, hence relative hrefs don't work. Previously we typically handled this problem by always using absolute hrefs in the entrypoint. Now that we're serving multiple builds on the same host, we need the entrypoint hrefs to point not at `/fragment.html`, but `/$build/fragment.html`. The server responds to entrypoint requests with the entrypoint from the build it thinks is best for the given user agent, which has already been transformed with the right `<base>` tag.

usergenic:
A fragment requested from the entrypoints would load but any assets it was requesting due to its now incorrect base href would fail to return.

aomarks:
Ah I see, yeah -- if the base tag was absolute, and the user intends it to point at build-specific resources, that would indeed not work. But... they should just use a relative base tag in that case though? It's not so clear what feature we could build, though, because how would we know whether any given href is supposed to point at the *real* root or the *build* root, since both are valid? The rules for the layout we're recommending now are:

- To reference resources shared across builds, use absolute hrefs (e.g. /images/foo.png, /api/launch-missile).
- To reference build-specific resources:
  - In your entrypoint, use relative URLs with `<base href="/">`. We'll update base for you.
  - Elsewhere, use relative URLs. You can use a `<base>` tag, but make sure it's relative.

usergenic:
As long as the caveat (maybe a note somewhere) is that absolute base hrefs will escape the served path prefix.  I guess that's not hard to grok.  Okay I am at peace with this.

src/base-tag-updater.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    const parsed = parse5.parse(contents);
+    let base = dom5.query(parsed, baseMatcher);
+    if (!base || dom5.getAttribute(base, 'href') === this.newHref) {
+      callback(null, file);
+      return;
+    }
+
+    dom5.setAttribute(base, 'href', this.newHref);
+    const updatedFile = file.clone();
```
usergenic: Consider calling the following:
```js
dom5.removeFakeRootElements(parsed);
```
before serializing to prevent generating ghost tags.

Not really that important, but maybe it would mess with someone's build if they expected we didn't add tags that weren't there.

aomarks: Done.

aomarks:
@usergenic Added a paragraph to the README to clarify that fragments won't be touched, and why.

Also, I'm writing up this general app layout pattern, so once we have that documentation we can link to it from here to add even more context.

README.md
```diff
@@ -240,6 +240,31 @@ const bundledBuildStream = forkStream(buildStream)
...
+
+const unbundledBuildStream = forkStream(buildStream)
+  // This build will be served from http://example.com/unbundled/
+  .updateBaseTag('/unbundled/')
+  .pipe(gulp.dest('build/unbundled'));
+
+const bundledBuildStream = forkStream(buildStream)
+  .pipe(project.bundler())
+  // While this build will be served from http://example.com/bundled/
+  .updateBaseTag('/bundled/')
```
usergenic: .pipe(project.updateBaseTag('/bundled/'))

README.md
```diff
@@ -240,6 +240,31 @@ const bundledBuildStream = forkStream(buildStream)
...
+```js
+const gulp = require('gulp');
+const mergeStream = require('merge-stream');
+const forkStream = require('polymer-build').forkStream;
+
+const buildStream = mergeStream(project.sources(), project.dependencies());
+
+const unbundledBuildStream = forkStream(buildStream)
+  // This build will be served from http://example.com/unbundled/
+  .updateBaseTag('/unbundled/')
```
usergenic: .pipe(project.updateBaseTag('/unbundled/'))

aomarks: Fixed on master.

README.md
aomarks: Fixed on master.

PR #208, opened 2017-05-12
Updated bundler to filter out all inlined files from its output.

usergenic:
- Bundler stream no longer emits any CSS or Javascript files which have been inlined into bundles.


src/test/bundle_test.ts
```diff
@@ -291,12 +291,6 @@ suite('BuildBundler', () => {
         {
           root: path.resolve('test-fixtures/bundle-project'),
           entrypoint: 'index.html',
-          sources: [
```
rictic: what's up here? I guess the idea is that the entrypoint is enough to get the rest of the files?

src/test/bundle_test.ts
```diff
@@ -383,16 +365,45 @@ suite('BuildBundler', () => {
     assert.include(bundledHtml, '.simply-red', 'simple-style.css');
   });
 
-  test('bundler does not output imports which have been bundled', async() => {
+  test(
+      'bundler does not output html imports which have been bundled',
+      async() => {
+        await setupTest({
+          entrypoint: 'entrypoint-only.html',
+          sources: ['framework.html', 'entrypoint-only.html'],
```
rictic: Are these sources necessary, given that entrypoint-only.html depends on framework.html?

Or are you adding them just to be doubly sure that framework.html was present but then was removed?

src/test/bundle_test.ts
usergenic: I just removed them because it turned out the sources bit of the config had no bearing on the test and it was misleading.  It was incidental and opportunistic, but not material.

src/test/bundle_test.ts
usergenic: I actually didn't change this test at all, just its name, to be similar to the new test for 'bundler does not output inlined scripts or styles'.  The diff didn't make that obvious.  Fixing that.

I went ahead and removed all other `sources:` entries from the configs, because they were all irrelevant as far as the bundler was concerned.  They really only effected the BuildAnalyzer, none of that mattered to the bundle tests.

Changes pushed.

src/test/bundle_test.ts
usergenic: But yes, the entrypoint and/or fragments is all that bundler needs.

PR #210, opened 2017-05-12
Use AsyncTransformStream rather than stream.Transform

rictic:
This lets us write our transform streams using async iterators, making them easier to read and understand, as well as consistently propagating errors.

 - [x] CHANGELOG.md not updated, internal change.


rictic:
(and yes, the formatting is kinda unfortunate. filed as https://github.com/angular/clang-format/issues/44 )

gulpfile.js
```diff
@@ -57,19 +57,23 @@ gulp.task('tslint', function() {
...
+      // automatically loaded by TypeScript during build. depcheck can't
+      // detect this so we ignore them here.
+
+      '@types/*',
+      // Also it can't yet parse files that use async iteration.
+      // TODO(rictic): remove these
+      'mz', 'multipipe', 'polymer-bundler',
+  ]}).then((result) => {
+    let invalidFiles = Object.keys(result.invalidFiles) || [];
+    let invalidJsFiles = invalidFiles.filter((f) => f.endsWith('.js'));
```
usergenic: const these i think

src/streams.ts
```diff
@@ -50,25 +54,171 @@ export function compose(streams: NodeJS.ReadWriteStream[]) {
...
+      const transformDonePromise = (async() => {
+        for
+          await(const value of this._transformIter(this._inputs)) {
+            // TODO(rictic): if `this.push` returns false, should we wait until
+            //     we get a drain event to keep iterating?
+            this.push(value);
+          }
+      })();
+      // TODO(rictic): blindly drain the rest of the inputs if _transformIter
+      //     returns early?
```
usergenic: Maybe we provide an explicit method to drain remaining inputs and throw an error/reject the promise if the inputs are not empty (i.e. if its closed?) as a guard against accidental early exits?  Also, why don't expose closed state on the AsyncQueue even as a readonly public getter?

```typescript
if (!this._inputs.closed) {
  this._writingFinished.reject(new Error('Unfinished inputs queue.'));
} else {
  this._writingFinished.resolve(undefined);
}
```

rictic: +1. it took some fiddling, but I think I got this working, + added a test

src/streams.ts
```diff
@@ -50,25 +54,191 @@ export function compose(streams: NodeJS.ReadWriteStream[]) {
...
+      transformDonePromise.then(() => {
+        if (this._inputs.finished) {
+          this._writingFinished.resolve(undefined);
+        } else {
+          this.emit(
+              'error',
+              new Error(
+                  `${this.constructor.name}` +
+                  ` did not consume all input while transforming.`));
+          this.push(null);
```
usergenic: What's this push(null) about?

rictic: It signals the end of the stream, that no more output will be written.

PR #212, opened 2017-05-13
Use build basePath as replacePrefix for service worker generation.

aomarks:
Part of the differential serving story.


src/test/service-worker_test.ts
```diff
@@ -185,6 +185,34 @@ suite('service-worker', () => {
...
+            buildRoot: testBuildRoot,
+            basePath: '/my/base/path'
+          })
+          .then((swFile: Buffer) => {
+            const fileContents = swFile.toString();
+            assert.include(fileContents, '"/my/base/path/index.html"');
+          });
+    });
+
+    test('basePath prefixes should not have double delimiters', () => {
```
usergenic: I think all these tests could be redone as async since generateServiceWorker is async
```typescript
test('...', async() => {
  const swFile = await serviceWorker.generateServiceWorker(...);
  const fileCOntents = swFile.toString();
  assert.include(...);
  ...
});
```

usergenic: But I see you were just following the pre-existing paradigm in the tests, so that's not a dealbreaker on this PR.

Add a note to top of file?

src/service-worker.ts
```diff
@@ -115,8 +116,26 @@ export async function generateServiceWorker(options: AddServiceWorkerOptions):
...
+
+  if (options.basePath) {
+    let replacePrefix = options.basePath;
+    // Avoid a "/foo//bar" situation.
+    if (replacePrefix.endsWith('/') &&
+        !swPrecacheConfig.stripPrefix.endsWith('/')) {
+      replacePrefix = replacePrefix.slice(0, -1);
+    }
+    if (swPrecacheConfig.replacePrefix) {
+      console.warn(
```
usergenic: The semantic of warn seems strange given this warning only occurs because the option is provided.  I guess its warning because this is the kind of thing that might trip people up?

usergenic:
(also obv have to resolve conflicts)

src/service-worker.ts
aomarks: Done. Yeah I agree, it's just something people should be aware of, so not a warning.

src/test/service-worker_test.ts
aomarks: Done.

PR #213, opened 2017-05-15
Prepare release

usergenic:
- Updated `polymer-analyzer` dependency to `^2.0.0` now that it is out of alpha.


package.json
```diff
@@ -30,8 +30,8 @@
     "multipipe": "^1.0.2",
     "parse5": "^2.2.2",
     "plylog": "^0.5.0",
-    "polymer-analyzer": "2.0.0-alpha.42",
-    "polymer-bundler": "^2.0.0-pre.17",
+    "polymer-analyzer": "^2.0.0",
+    "polymer-bundler": "2.0.0-pre.18",
```
stramel: Should this wait for the full 2.0 release? just wondering

usergenic: bundler is not quite 2.0.0 -- probably the next release after a couple deep import paths are fixed etc.

PR #222, opened 2017-05-18
Normalize push manifest URLs to be relative, and fix basePath join bug.

aomarks:
Also renamed some parameters to be more clear.

Fixes #746.

 - [x] CHANGELOG.md updated


src/push-manifest.ts
```diff
@@ -209,18 +211,26 @@ export class AddPushManifest extends AsyncTransformStream<File, File> {
...
+    //
+    // TODO Decide whether they should really be relative or absolute. Relative
+    // was chosen here only because most links were already relative so it was
+    // a smaller change, but
+    // https://github.com/GoogleChrome/http2-push-manifest actually shows
+    // relative for the keys and absolute for the values.
+    const normalize = (p: string) =>
+        path.posix.join(this.basePath, p).replace(/^\/+/, '');
+
+    const normalized: any = {};
```
rictic: can we say more about these types than `any`?

aomarks: I have defined a push manifest type in prpl-server, but don't have it here. It's the same type it already has above, so I wasn't going to address it for now.

rictic: I'd say copy-paste the push manifest type in from prpl-server here for now.

aomarks: Actually, turns out there was a type defined! We just weren't using it for some reason.

PR #230, opened 2017-05-23
Generate relative URLs for service worker pre-cached assets.

aomarks:
See https://github.com/Polymer/polymer-cli/issues/761


src/service-worker.ts
```diff
@@ -115,14 +115,17 @@ export async function generateServiceWorker(options: AddServiceWorkerOptions):
   // swPrecache will determine the right urls by stripping buildRoot.
   // NOTE:(usergenic) sw-precache generate() apparently replaces the
   // prefix on an already posixified version of the path on win32.
-  swPrecacheConfig.stripPrefix = posixifyPath(buildRoot);
+  //
+  // We include a trailing slash so that we end up with relative URLs instead
+  // of absolute. This is needed for builds mounted at non-root paths. Note
+  // that the service worker fetches relative to its own URL.
+  swPrecacheConfig.stripPrefix = addTrailingSlash(posixifyPath(buildRoot));
```
usergenic: This comment threw me off because it wasn't clear to me why a URL without a trailing slash results in an absolute URL.  I figured it out but was thinking maybe "We include a trailing slash in so that URLs stripped of this prefix will be without a leading slash, i.e. relative instead of absolute."

Too much?

aomarks: Yeah, I can see that's confusing! Updated the comment, PTAL?

usergenic: 👍  Thank you :)

PR #235, opened 2017-06-01
bug(build): Use polymer root as cwd to fix build output directory

hvdb:
The stream containing the sources and dependencies now have the paths relative to the polymer.root instead of the process.CWD.

Breaking change.

The build output is now correctly written to the build folder.

As there are multiple ways into fixing this, please let me know if the solution i made is not according to your standards!

I will do some more testing, my application is now properly outputted in the correct build folder.
But i do see one error (webcomponentsjs) is still inside the wrong folder.
Will look into that one and test with more apps.

closes: https://github.com/Polymer/polymer-build/issues/234

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -105,6 +105,7 @@ export class BuildAnalyzer {
   analyzer: Analyzer;
   started: boolean = false;
   sourceFilesLoaded: boolean = false;
+  cwd: string;
 
   private _sourcesStream: NodeJS.ReadableStream;
   private _sourcesProcessingStream: NodeJS.ReadWriteStream;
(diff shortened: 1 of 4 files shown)


hvdb:
Ok, all done for me.

Tested it with and without a root path.
If no root path is specified it works just like before.

If root is specified the build output is no longer put inside a subfolder with the name of the root folder.

as far as i can see everything works.
Tested it with our application and looked at the before and after result, they are the same (without the sub dir ofcourse)

Please let me know!

hvdb:
Any update?

hvdb:
Anyone?

hvdb:
@usergenic @rictic 

Can someone please look at this?
It blocks our build process currently.

justinfagnani:
Hi @hvdb thanks for your patience with this one.

We're looking at it to see if this is the correct behavior. I'm a little worried that different people have different expectations for how `root` should behave, and this change would break some users. Maybe there's a way to do this with another option that's not breaking.

Either way we definitely need some tests that fail before this PR and pass with it. That'll also help review the intended behavior.

aomarks:
Closing this pull request because this repo is now archived. If this PR is still active, please submit a new one at https://github.com/Polymer/tools.

PR #236, opened 2017-06-02
Dont prefetch direct imports

usergenic:
 - Because adding prefetch links for links already in the page is actually counter-productive (incurs additional cost, like browsers fetching something 3 times instead of 1 in certain cases!)


src/prefetch-links.ts
```diff
@@ -67,25 +67,40 @@ export class AddPrefetchLinks extends AsyncTransformStream<File, File> {
...
-                           })].filter((d) => !d.lazy);
+
+      const allDependencyUrls = [
+        ...document.getFeatures({
+          kind: 'import',
+          externalPackages: true,
+          imported: true,
+          noLazyImports: true
+        })
+      ].filter((d) => !d.lazy).map((d) => d.document.url.replace(/^\//, ''));
```
aomarks: Add a comment about why you strip the leading `/`?

usergenic: I totes didn't need to strip these.  It was cargo'd paranoia from the earlier version of the code.  That code has now been removed.

PR #238, opened 2017-06-04
Remove `type` for custom-elements-es5-adapter.js

baocang:
In HTML 5, the type attribute is optional and defaults to `text/javascript`


FredKSchott:
We'll keep it for now to be explicit, thanks for the contribution @baocang!

PR #248, opened 2017-06-28
Treat lazy-imports encountered as fragments when producing the push-manifest

usergenic:
- If your polymer.json doesn't specify fragments, that's okay as long as they're lazy-imports of the shell or other explicit graph entrypoint.
 - Fixes https://github.com/Polymer/polymer-build/issues/247


usergenic:
cc @graynorton

usergenic:
Note there's an implicit dependency on https://github.com/Polymer/polymer-bundler/pull/572 in that the changes to code in this PR removed an uncovered-by-tests hack for working around/ignoring the dependency on the shell from other files (which happens post bundle).  When that PR is makes it into 2.2.0 bundler release, polymer-build's package.json will be updated to use that version of bundler and push manifest will continue to *not* push the shell along with fragments.

usergenic:
polymer-bundler 2.2.0 published 🎉

Have updated and pushed package.json

PR #249, opened 2017-06-29
Prepare to release polymer-build v1.6.0

usergenic:
## [1.6.0] - 2017-06-29
* Automatically recognize any lazy-imports encountered as fragments when generating push-manifest.
* The `addPushManifest` feature now honors the laziness of html-imports and excludes them from the set of their importers' pushed assets.
* Upgraded Polymer Bundler to 2.2.0, which updated the shell strategy so that the shell is no longer linked to from other bundles. See [Bundler issue #471](https://github.com/Polymer/polymer-bundler/issues/471) for more details.
* [x] CHANGELOG.md has been updated


usergenic:
Yeah I don't think any docs need to change yet on this stuff.  I built shop with it and everything seemed happy too.

PR #262, opened 2017-08-16
Remove unnecessary casts and TODO's

TimvdLippe:
I found these TODO's while looking through the code. Since the corresponding PR has since been closed and the typings of Node are updated, the casts can be removed.

 - [x] CHANGELOG.md has not been updated, internal NOOP change


src/html-splitter.ts
```diff
@@ -102,14 +102,17 @@ export class SplitFile {
...
+    console.assert(
+        this.parts.get(path) !== undefined,
+        `Trying to save unexpected file part "${path}".`);
+    console.assert(
+        this.parts.get(path) === null,
+        `Trying to save already-saved file part "${path}".`);
+    console.assert(
+        this.outstandingPartCount > 0,
+        `Trying to save valid file part "${path}", ` +
+            `but somehow no file parts are outstanding.`);
```
stramel: Can't template literal strings span multiple lines without the need for concatenation

stramel:
Dang you beat me to these! Meant to throw out a PR when I was working on the multiple builds for custom-build template

src/html-splitter.ts
TimvdLippe: That would insert new lines instead and in this case, we want to assert that the single line is a warning.

TimvdLippe:
Thanks for the speedy reviews!

PR #275, opened 2017-11-29
Use ResolvedUrl type in loader.

rictic:
- [x] CHANGELOG.md not updated, internal change


rictic:
Also fixed a legit bug from https://github.com/Polymer/polymer-analyzer/pull/734

usergenic:
Thanks for updating to skip the back-reference import!

usergenic:
While I feel like the html-script-back-reference check should be in the filtering as opposed to in the createPushEntryFromImport function, that's subjective enough of a feels that this is still LGTM

rictic:
+1, moved the filtering into one place.

PR #276, opened 2017-12-08
fix silent (!) build failure when several source filepaths cannot be found

vlukashov:
This PR ensures that the [`scan()`](https://github.com/Polymer/polymer-analyzer/blob/master/src/core/analysis-context.ts#L337) method _gets completed_ for every source file in the project. Currently, the scan method _gets started_ but would never _get completed_ if there is more than one missing source file dependency in dependencies tree.

A `scan()` of source file gets completed when all dependencies of that file are known (resolved or rejected). That's why when the project's sources stream is completed it's important to ensure _all_ deferred source get rejected--not only the first one.

If some of the deferred dependencies remain pending, the analysis of the toplevel source file never continues after [`await this._cache.dependencyGraph.whenReady(resolvedUrl);`](https://github.com/Polymer/polymer-analyzer/blob/master/src/core/analysis-context.ts#L355). 

This PR fixes https://github.com/Polymer/polymer-cli/issues/936, https://github.com/Polymer/polymer-cli/issues/904 and https://github.com/Polymer/polymer-build/issues/251.

It might be also related to https://github.com/Polymer/polymer-build/issues/272.

### DIFF
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -249,7 +249,6 @@ export class BuildAnalyzer {
       if (this.config.isSource(filePath)) {
         const err = new Error(`Not found: ${filePath}`);
         this.loader.rejectDeferredFile(filePath, err);
-        return;
       }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 <!-- Add new, unreleased changes here. -->
+* Fixed issue where the build silently fails when several source dependencies are missing
 
 ## [2.1.1] - 2017-10-23
 * Updated `polymer-bundler` to 3.1.1, to fix an issue with deprecated CSS imports being inlined into the wrong templates.
--- a/src/test/analyzer_test.ts
+++ b/src/test/analyzer_test.ts
@@ -182,7 +182,7 @@ suite('Analyzer', () => {
 
         let errorCounter = 0;
         const errorListener = (err: Error) => {
-          assert.equal(err.message, '2 error(s) occurred during build.');
+          assert.equal(err.message, '3 error(s) occurred during build.');
           errorCounter++;
           if (errorCounter >= 2) {
             done();
--- a/src/test/analyzer_test.ts
+++ b/src/test/analyzer_test.ts
@@ -182,7 +182,7 @@ suite('Analyzer', () => {
 
         let errorCounter = 0;
         const errorListener = (err: Error) => {
-          assert.equal(err.message, '1 error(s) occurred during build.');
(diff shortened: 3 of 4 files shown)


jsilvermist:
Should there also be a test for a broken file path in a lazy import added as well? I believe the lazy import caused silent failure even if there was only 1 broken path, not needing 2 before it happened.

vlukashov:
It would be a good idea to investigate if Polymer/polymer-cli#904 would be fixed by the same change. I have not done that (and that's why there is no test case for lazy imports).

Would you like to contribute? Here is how:
 - create a fork of the polymer-build repo
 - add my fork as a remote (`git remote add vlukashov https://github.com/vlukashov/polymer-build.git`)
 - create a branch based on this PR (`git checkout fix/251-silent-build-failures && git checkout -b your-own-branch-name`)
 - add a test case for lazy imports
 - push the branch to your fork and create a pull request targeting either `Polymer:master` or `vlukashov:fix/251-silent-build-failures`

jsilvermist:
Tested locally and seems to be fixed.

Created a PR for a commit adding the `lazy-import` test here:
vlukashov/polymer-build#1

vlukashov:
@jsilvermist nice! I've just added your test case to the PR.

But now it looks like the Google CLA bot wants you to write here a comment explicitly stating that you do accept the CLA for this repo.

jsilvermist:
Nah I've signed the CLA don't worry, it's just how the google bot acts if anyone adds a commit from a different author to a PR.

keithbloom:
Hi

I have a build that is silently failing with a specific set of incorrect links

1. A component with a `<link rel="import">` to a component that is missing
2. Two components that have `<link rel="stylesheet"`> where the stylesheet is missing

This configuration caused Polymer Build to fail silently. When I applied the fix in this PR the build fail correctly and echos the messages from Analyzer to the console.

The sooner this PR can be merged and released the better. This bug means the build process can fail without any indication of the problem.

src/analyzer.ts
```diff
@@ -249,7 +249,6 @@ export class BuildAnalyzer {
       if (this.config.isSource(filePath)) {
         const err = new Error(`Not found: ${filePath}`);
         this.loader.rejectDeferredFile(filePath, err);
-        return;
```
justinfagnani: This early return is unintuitive. It seems dangerous to not process the remaining errors or close the dependencies steam. I also don't see how the test should show more errors with this addition. Does the modified test fail without this change?

keithbloom: I think the change removes the early return.

justinfagnani: Right, so I shouldn't do reviews on my phone on the subway :)

justinfagnani:
Thanks for the PR @vlukashov!

And thanks for the correction on the review @keithbloom :)

PR #279, opened 2017-12-28
Add autoBasePath

abdonrd:
This new flag sets the entrypoint `<base>` tag for all builds to match the name of that build.

Depends on https://github.com/Polymer/polymer-project-config/pull/48.


aomarks:
To fix the TypeScript error, we'll need to declare the option in https://github.com/Polymer/polymer-project-config/blob/master/src/index.ts#L126. Feel free to send a PR for that, and I'll get it released so that it will then be available here.

aomarks:
> To fix the TypeScript error, we'll need to declare the option in https://github.com/Polymer/polymer-project-config/blob/master/src/index.ts#L126. Feel free to send a PR for that, and I'll get it released so that it will then be available here.

Oh, oops, nevermind. I missed that you already sent the other PR!

abdonrd:
Closed in favor to https://github.com/Polymer/polymer-project-config/pull/48

PR #254, opened 2017-07-21
Allow filtering analysis warnings

nelsonsilva:
This is a quick fix for #152 which reuses linting options for `polymer-cli` to allow filtering warnings emitted during the analysis.

PR #289, opened 2018-02-23
prepare release 2.2.0

bicknellr:


bicknellr:
Needs to be updated since #254 was merged.

PR #291, opened 2018-03-07
Move some build things into build

aomarks:
- optimize-streams from Polymer CLI
- babel-plugin-bare-specifiers from Polyserve

No code changes, purely moving modules from other repos. Discussed offline with justinfagani@


justinfagnani:
It'd be great to port some tests over, especially for the specifier plugin. The current tests in polyserve are integration tests with the server, but some unit tests should be relatively easy with some babel infrastructure.

PR #292, opened 2018-03-08
Simplfy GenericOptimizeTransform interface.

aomarks:
No longer takes an `any` options to pass to transform function. Instead, just assumes the subclass will handle options itself.

This is motivated by an upcoming change, where the `JSCompileTransform` is going to need access to the `File` object.

Got rid of `JSBabelTransform` class because it was now doing almost nothing.

### DIFF
--- a/src/optimize-streams.ts
+++ b/src/optimize-streams.ts
@@ -12,7 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {transform as babelTransform, TransformOptions as BabelTransformOptions} from 'babel-core';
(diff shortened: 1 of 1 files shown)


rictic:
Are there tests of these classes?

aomarks:
> Are there tests of these classes?

Only integration tests in their previous homes, unfortunately.

rictic:
Hm. I *guess* it's not blocking, but it rather sucks to move code away from its tests.

aomarks:
@rictic PTAL, moved over the unit test from polymer-cli (and merged its streams.ts into the one here)

PR #293, opened 2018-03-08
build compile transformer now rewrites bare module specifiers to paths.

aomarks:
Also:

- Bare specifier rewriter now works on Windows.
- Bare specifier rewriter no longer tries to resolve fully qualified URLs.
- Added `moduleResolution` option to `OptimizeOptions`.
- Combined compile and minimize babel transforms.


justinfagnani:
We should add a new option to match polyserve and the bundler: `moduleResolution: 'node'|'none'`

aomarks:
> We should add a new option to match polyserve and the bundler: moduleResolution: 'node'|'none'

Done PTAL

src/babel-plugin-bare-specifiers.ts
```diff
@@ -15,6 +15,7 @@
 import {dirname, relative} from 'path';
 import resolve = require('resolve');
 import {NodePath} from 'babel-traverse';
+import * as url from 'url';
```
justinfagnani: use `whatwg-url`

src/babel-plugin-bare-specifiers.ts
```diff
@@ -52,6 +57,10 @@ export const resolveBareSpecifiers =
           let relativeSpecifierUrl =
               relative(dirname(filePath), resolvedSpecifier);
 
+          if (process.platform === 'win32') {
```
justinfagnani: consider using the is-windows package

src/optimize-streams.ts
```diff
@@ -119,17 +110,25 @@ export class GenericOptimizeTransform extends Transform {
...
 export class JSCompileTransform extends GenericOptimizeTransform {
-  constructor() {
-    const transformer = (content: string, file: File) =>
-        babelTransform(content, {
-          presets: [babelPresetES2015NoModules],
-          plugins: [
+  constructor(options: OptimizeOptions['js']) {
+    const transformer = (content: string, file: File) => {
+      const presets = [];
+      const plugins = [
```
justinfagnani: add comment explaining why we're loading sync plugins without the transform. Also, make these a single array to use with the minify transform.

src/optimize-streams.ts
```diff
@@ -143,8 +142,8 @@ export class JSMinifyTransform extends GenericOptimizeTransform {
         babelTransform(content, {
           presets: [minifyPreset(null, {simplifyComparisons: false})],
           plugins: [
-            babelPluginSyntaxObjectRestSpread,
-            babelPluginSyntaxDynamicImport,
+            objectRestSpreadSyntax,
```
justinfagnani: Add todo/issue to combine with JSCompileTransform

aomarks: Just went ahead and did it.

src/optimize-streams.ts

src/babel-plugin-bare-specifiers.ts
aomarks: Done.

src/babel-plugin-bare-specifiers.ts
aomarks: Done.

aomarks:
PTAL

justinfagnani:
LGMT still

PR #302, opened 2018-03-13
Prepare 2.3.1 release.

aomarks:
Also fix compilation error because JS transform option did not allow "none" setting.


bicknellr:
What happens if a user tries to build with this option as `none`?

aomarks:
> What happens if a user tries to build with this option as none?

Then no module name to path transformation happens. See https://github.com/Polymer/polymer-project-config/blob/master/src/index.ts#L198.

rictic:
Then we'll fail on any modules with bare module specifiers I'd guess

PR #303, opened 2018-03-13
Fix bug where compile/minify excluded were ignored.

aomarks:
My earlier refactor in https://github.com/Polymer/polymer-build/pull/292 was totally wrong with respect to filtering using the `compile.exclude` and `compile.minify` options.

Also do compile/minification in the right order (Babel plugins run first to last, but presets run last to first :/).


justinfagnani:
We should add a regression test

src/optimize-streams.ts
```diff
@@ -113,28 +114,37 @@ export class GenericOptimizeTransform extends Transform {
...
         presets.push(babelPresetES2015NoModules);
         plugins.push(
-            externalHelpersPlugin,
-            objectRestSpreadTransform,
+            babelPluginExternalHelpers,
+            babelTransformPluginObjectRestSpread,
         );
       }
-      if (options.moduleResolution === 'node') {
+      if (nodeResolution) {
```
justinfagnani: why not  `if (options.moduleResolution === 'node') {`, or ` if (options.moduleResolution !== 'node') {`? At least do an explicit check against undefined.

aomarks:
> We should add a regression test

Done. It is also covered by an integration test in polymer cli, but better to have one here.

src/optimize-streams.ts

PR #309, opened 2018-03-14
Workbox

### LINKED ISSUE #250: Switch from sw-precache to workbox
What about use [`workbox`](https://github.com/GoogleChrome/workbox) instead of [`sw-precache`](https://github.com/GoogleChrome/sw-precache) in `polymer-build` to generate service worker code?

prateekbh:
Fixes #250

Start using `[workbox](https://github.com/GoogleChrome/workbox)` instead of `sw-precache` for service worker generation.

I am not sure if I have to add a new release or add it somewhere or how. please help around the CHANGELOG


src/service-worker.ts
```diff
@@ -98,97 +86,76 @@ export async function generateServiceWorkerConfig(
...
-
-  // swPrecache will determine the right urls by stripping buildRoot.
-  // NOTE:(usergenic) sw-precache generate() apparently replaces the
-  // prefix on an already posixified version of the path on win32.
-  //
-  // We include a trailing slash in `stripPrefix` so that we remove leading
-  // slashes on the pre-cache asset URLs, hence producing relative URLs
-  // instead of absolute. We want relative URLs for builds mounted at non-root
-  // paths. Note that service worker fetches are relative to its own URL.
-  swPrecacheConfig.stripPrefix = addTrailingSlash(posixifyPath(buildRoot));
```
prateekbh: this is something i am missing on right now, and would probably add as another manifestTransform

src/service-worker.ts
```diff
@@ -98,97 +86,76 @@ export async function generateServiceWorkerConfig(
...
 
-  if (swPrecacheConfig.navigateFallback === undefined) {
+  if (workboxConfig.globDirectory === undefined) {
+    workboxConfig.globDirectory = buildRoot;
+  }
+
+  if (workboxConfig.importScripts === undefined) {
+    // Map all application routes to the entrypoint.
+    workboxConfig.importScripts =
+      ['https://storage.googleapis.com/workbox-cdn/releases/3.0.0/workbox-sw.js'];
```
jeffposnick: The `workbox-build` node module exports a `getModuleUrl()` function, which will return the current CDN URL. I'd switch to that instead of hardcoding.

https://github.com/GoogleChrome/workbox/blob/ed2a847baf34c17c0f019d11cdfb1464fe3a9e19/packages/workbox-build/src/index.js#L86

(This is another method that is currently only used internally in the Webpack plugin, but it is supported.)

src/service-worker.ts
```diff
@@ -98,97 +86,76 @@ export async function generateServiceWorkerConfig(
...
-      replacePrefix = replacePrefix + '/';
-    }
-    if (swPrecacheConfig.replacePrefix) {
-      console.info(
-          `Replacing service worker configuration's ` +
-          `replacePrefix option (${swPrecacheConfig.replacePrefix}) ` +
-          `with the build configuration's basePath (${replacePrefix}).`);
-    }
-    swPrecacheConfig.replacePrefix = replacePrefix;
+    workboxConfig.manifestTransforms = [(entries: ManifestEntry[]) => {
```
jeffposnick: I think you can get away with just setting

```js
workboxConfig.modifyUrlPrefix = {
  '': options.basePath
}
```

(Translation: replace the empty string as a prefix with the `options.basePath` value.)

src/service-worker.ts
```diff
@@ -98,97 +86,76 @@ export async function generateServiceWorkerConfig(
...
-            reject(err);
-          } else {
-            resolve(new Buffer(fileContents));
-          }
-        });
+  Promise<Buffer> {
+  const workboxConfig = await generateServiceWorkerConfig(options);
+  return await <Promise<Buffer>>(new Promise((resolve) => {
+    logger.debug(`writing service worker...`, workboxConfig);
+    generateSWString(workboxConfig).then(({ swString }) => {
```
jeffposnick: I think it's worth checking for `warnings` (which is an array of strings) in the `generateSWString()` response, and logging anything that's returned.

jeffposnick (changes requested):
Thanks for working on this, @prateekbh!

I'll let the maintainers weigh in, but here are my comments from the Workbox perspective.

src/service-worker.ts
```diff
@@ -98,97 +86,76 @@ export async function generateServiceWorkerConfig(
...
   const project = options.project;
   const buildRoot = options.buildRoot;
-  const swPrecacheConfig: SWConfig =
-      Object.assign({}, options.swPrecacheConfig);
+  const workboxConfig: WorkboxConfig =
+    Object.assign({}, options.workboxConfig);
 
   const depsIndex = await project.analyzer.analyzeDependencies;
-  let staticFileGlobs = Array.from(swPrecacheConfig.staticFileGlobs || []);
+  let staticFileGlobs = Array.from(workboxConfig.globPatterns || []);
```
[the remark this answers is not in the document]
prateekbh: So I saw a concept of `bundled/unbundled` asset, which are passed to this method asn I just use them as exact globs to match and construct the precacheManifest

src/service-worker.ts
```diff
@@ -98,97 +86,76 @@ export async function generateServiceWorkerConfig(
...
   staticFileGlobs = staticFileGlobs.map((filePath: string) => {
     if (filePath.startsWith(project.config.root)) {
       filePath = filePath.substring(project.config.root.length);
     }
-    return path.join(buildRoot, filePath);
+    return removeLeadingSlash(filePath);
   });
 
-  if (swPrecacheConfig.navigateFallback === undefined) {
+  if (workboxConfig.globDirectory === undefined) {
```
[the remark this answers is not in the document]
prateekbh: Done

src/service-worker.ts
prateekbh: TIL! Done

src/service-worker.ts
prateekbh: Done

src/service-worker.ts
prateekbh: Done

justinfagnani:
Thanks @prateekbh, this is really helpful! I'll review this ASAP.

src/service-worker.ts
```diff
@@ -98,97 +86,61 @@ export async function generateServiceWorkerConfig(
   options = Object.assign({}, options);
   const project = options.project;
   const buildRoot = options.buildRoot;
-  const swPrecacheConfig: SWConfig =
-      Object.assign({}, options.swPrecacheConfig);
+  const workboxConfig: WorkboxConfig =
+    Object.assign({}, {
+      globDirectory: buildRoot,
+      importScripts: [getModuleUrl('workbox-sw')],
```
jeffposnick: Thanks for moving this into the `Object.assign()` logic.

In the back of my head, I'm wondering whether something special needs to be done for `importScripts`, apart from just overwriting the default `importScripts` array (which contains the Workbox CDN URL).

For instance, if `options.workboxConfig.importScripts === ['my-push-hander.js']`, I think the developer's intention is that they still want Workbox from the CDN, but they also want to load in their custom push handler script. With the code as it is right now, they'd just end up with the push handler and no Workbox CDN.

If we assume that the developer will *always* load Workbox from the CDN URL, then instead of overwriting the default `importScripts` in this `Object.assign()`, you could instead do

```js
workbox.config.importScripts =[getModuleUrl('workbox-sw')];
if ('importScripts' in options. workboxConfig) {
  workbox.config.importScripts = workbox.config.importScripts.concat(
    options.workboxConfig.importScripts);
}
```

Or something like that. It might not perfectly capture all use cases, but I think it gives some more flexibility—and if we get folks who ask for non-CDN builds, that would involve re-evaluating this anyway.

jeffposnick (changes requested):

src/service-worker.ts
```diff
@@ -98,97 +86,63 @@ export async function generateServiceWorkerConfig(
...
   });
 
-  if (swPrecacheConfig.navigateFallback === undefined) {
-    // Map all application routes to the entrypoint.
-    swPrecacheConfig.navigateFallback =
-        path.relative(project.config.root, project.config.entrypoint);
-  }
+  // Add the workbox cdn scripts
+  workboxConfig.importScripts = workboxConfig.importScripts || [];
+  workboxConfig.importScripts.push(getModuleUrl('workbox-sw'));
```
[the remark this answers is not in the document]
prateekbh: ah yes, makes perfect sense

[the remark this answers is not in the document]
prateekbh: Done!

justinfagnani:
@prateekbh thanks again for all the work you've put in here. We've been quite busy working on Polymer 3.0 support in tools, which is the only reason why I haven't ben able to review this yet.

Since this PR was opened, we've moved to a mono at Polymer/tools. We'll need to port this PR over there.

@rictic, @graynorton and @kevinpschaaf putting this on your radar to look at if 3.0 issues are ironed out in the next week or so. This is a breaking change because the user-provided configs are different from sw-precache. We need to decide if we can support both workbox and sw-precache, or roll this into a set of CLI 2.0 breaking changes.

prateekbh:
@rictic, @graynorton and @kevinpschaaf monorepo is kinda huge can you point me the location where the sw generation location exists.

merlinnot:
https://github.com/Polymer/tools/blob/master/packages/build/src/service-worker.ts

Issue #269, opened 2017-10-30
Support transpilation of ES modules (both inline and external scripts)


usergenic:
Right now polyserve is the way to run/test code using `<script type=module>` in browsers without module support, because it converts the HTML and JS code on the fly based on browser capabilities.  We need to support the same conversions in `polymer-build`.

To the extent possible we should generalize and extract the transformation code out of polyserve and using the same code in both places.

daKmoR:
any updates on this? would be awesome to be able to use `<script type=module>` in production :)

usergenic:
@daKmoR This is on our short-term roadmap but work hasn't actually started on this item yet.  Aiming for  December.

daKmoR:
sorry for being eager for information but and has there been any process on this?

PR #311, opened 2018-03-15
Factor out JS transformation into its own module.

aomarks:
Adds a `jsTransform` function, factored out of `optimize-streams`.

This is part of unifying the compilation logic between build and serve, and is
part of #269.


aomarks:
> It might be good to name jsTransform something that indicates that it returns a transformer.

It doesn't, it takes a string and returns a string.

PR #312, opened 2018-03-16
Integrate module to AMD babel plugin to jsTransform.

aomarks:
Also:
- Add unit tests to `jsTransform`.
- Expose `isComponentRequest` option for Polyserve to use.
- Renamed `compile` to `compileToEs5` to clarify what it does.

Part of #269.

(Note this does not yet include the dependency chaining from https://github.com/Polymer/polyserve/commit/fc33cd50bfa87bf187be797ff7c336dece05c95c which I intend to add in a follow up).

### DIFF
--- a/src/test/js-transform_test.ts
+++ b/src/test/js-transform_test.ts
@@ -1,0 +1,103 @@
+/**
+ * @license
+ * Copyright (c) 2018 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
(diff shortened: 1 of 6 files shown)


src/test/js-transform_test.ts
```diff
@@ -0,0 +1,103 @@
...
+      import { dep1 } from 'dep1';
+      import { dep2 } from 'dep2';
+      import { dep2A } from 'dep2/a';
+      import { dep3 } from 'dep3';
+      import { dep4 } from 'dep4';
+
+      import { p1 } from '/already/a/path.js';
+      import { p2 } from './already/a/path.js';
+      import { p3 } from '../already/a/path.js';
+      import { p4 } from '../already/a/path.js';
```
justinfagnani: Why the duplication here? What's it testing?

aomarks: Since the refactoring, I wanted to associate tests for jsTransform features directly in a jsTransform test, instead of via optimize-streams, because this function will now be shared with Polyserve and stands alone. I removed some edge cases from the optimize-stream test, since now we just care that the transformation ran at all.

PR #313, opened 2018-03-16
Support all ES2016, ES2017 plugins and some ES2018.

aomarks:
Adds:
- exponentiation (ES2016)
- async/await (ES2017)
- async generators (ES2018)

Also improves tests and documentation.

Supersedes https://github.com/Polymer/polyserve/pull/238

cc @LarsDenBakker


stramel:
Hang on isn't this not needed if we go with babel-preset-env?

justinfagnani:
@stramel maybe not, but we're not sure yet how we want to configure preset-env so no transform is required in certain browsers. If this is temporary until then that's fine.

aomarks:
> Hang on isn't this not needed if we go with babel-preset-env?

Yes, but we're not on babel-preset-env yet, just wanted to throw this in quickly.

My open question about babel-preset-env is how to handle the minify-but-not-transform case. I'm not sure how to get babel-preset-env to give us all the syntax plugins so that babel doesn't blow up on new syntax, without also getting the transpilation plugins. LMK if you know away, otherwise will investigate later.

src/js-transform.ts
```diff
@@ -24,14 +24,17 @@ import {resolveBareSpecifiers} from './babel-plugin-bare-specifiers';
 const babelSyntaxPlugins = [
   require('babel-plugin-syntax-dynamic-import'),
   require('babel-plugin-syntax-object-rest-spread'),
+  require('babel-plugin-syntax-async-functions'),
+];
+const babelCompilePlugins = [
+  require('babel-plugin-transform-object-rest-spread'),
```
justinfagnani: is this now everything from ES2018? Did we get all of ES2016&7 too? Why not change the below to use `babel-preset-es2015`?

aomarks: > is this now everything from ES2018? Did we get all of ES2016&7 too?

I can go through and find any missing ones. We don't have exponentiation operator, for example. But it does feel like we should just do preset env. I just sent this PR because we had some specific requests for these features.

> Why not change the below to use babel-preset-es2015?

Not sure I understand. We so use babel-preset-es2015, but only during transform. All the babel-preset-es* plugins do transformations. Similar problem we have with preset env. So we have to extract all the specific syntax plugins anyway, for the minimize-but-not-compile case.

justinfagnani: But this is the transform and not just syntax plugin, yes? So if we're adding this to the list of transforms, couldn't we just update the preset below to use `babel-preset-es2015`?

Or maybe you meant to add the syntax plugin here?

Sorry if I'm confused here too, I keep reading syntax or transform and thinking the other... :(

aomarks: Oh, you mean update preset-es2015 to preset-es2017. Yes, let's just do that, then we don't need to list the specific transform plugins. We do still need to list the specific syntax plugins, though.

justinfagnani: preset-es2018 if we're talking about object spread

aomarks: There is no preset-es2018, they are still under "experimental": http://babeljs.io/docs/plugins/#experimental

justinfagnani: Also, I guess we need to be sure what syntax plugins are required to get up to ES2018. I suppose Babylon natively supports everything up to 2015?

aomarks: Yep, so no plugins give you all the 2015 syntax. The 2015 preset adds transforms for those features. Then 2016 and 2017 add syntaxes and transforms for those years.

justinfagnani: https://www.npmjs.com/package/babel-preset-es2018 🤔

justinfagnani: ^ no README, no repo link, barely any downloads. Ok, don't use that :)

aomarks: Looked at that. It's just there to prevent squatting I assume: https://unpkg.com/babel-preset-es2018@1.0.0/

justinfagnani: ahhh... I originally just saw hzoo and was like "cool, use it"

aomarks: Ok, PTAL.

I added more documentation, a couple plugins (exponentiation and async generators), and organized the plugins by year.

All of the ES2018 features are not yet available, e.g. regexp named capture groups is still in development: https://github.com/babel/babel/pull/7105.

aomarks: Oh also, decided not to use the ES2017 preset, because turns out these presets don't include their prior year plugins. Since ES2016 and ES2017 only add one plugin each, it seemed simpler to just list them (since we have to list the syntax plugins separately anyway) than to merge the presets.

shawncplus:
Perhaps I'm missing something but because this doesn't update the gulpfile to uncomment the appropriate babel helpers you can't really use any of the ES2017+ features after the build since the helpers won't be injected into `lib/babel-helpers.min.js and therefor as part of `polymer-cli` won't be injected into the entrypoint

aomarks:
Which gulpfile?

shawncplus:
https://github.com/Polymer/polymer-build/blob/master/gulpfile.js

aomarks:
Oh, gross, I didn't realize that's how we generated the helpers. Need to think about that.

PR #317, opened 2018-03-20
Fix bare specifier rewrites for top-level-to-dependency imports

justinfagnani:


src/babel-plugin-bare-specifiers.ts
```diff
@@ -29,61 +30,71 @@ type HasSpecifier =
...
                   packageJson['jsnext:main'] || packageJson.main;
               return packageJson;
             },
-          });
+      });
 
-          let relativeSpecifierUrl =
-              relative(dirname(filePath), resolvedSpecifier);
+      let relativeSpecifierUrl: string;
+      // If we have an import from that cross the top-level / componentDir
```
aomarks: from that

src/babel-plugin-bare-specifiers.ts
```diff
@@ -29,61 +30,71 @@ type HasSpecifier =
...
+      // boundary, fix up the path to generate polyserve-style ../ cross-package
+      // imports.
+      if (isComponentRequest && componentDir !== undefined &&
+          rootDir !== undefined && resolvedSpecifier.startsWith(componentDir) &&
```
aomarks: I think startsWith isn't safe for this check unless we ensure there's a trailing slash

src/babel-plugin-bare-specifiers.ts
justinfagnani: fixed

src/babel-plugin-bare-specifiers.ts
justinfagnani: fixed

justinfagnani:
PTAL

PR #319, opened 2018-03-21
Replace `isComponentRequest` option with `componentRequestRelationship`

FredKSchott:
- Replace `isComponentRequest` option with `componentRequestRelationship`: To rewrite an import specifier across a "component" boundry, you must now set the compnent's relationship to that boundry. Normally this will be either `../` for unscoped packages and `../../` for scoped packages.
- Ran `npm run format` and pulled in some small formatting changes by accident.


src/babel-plugin-bare-specifiers.ts
```diff
@@ -34,7 +34,7 @@ type HasSpecifier =
  */
 export const resolveBareSpecifiers = (
     filePath: string,
-    isComponentRequest: boolean,
+    componentRequestRelationship: string,
```
justinfagnani: I think passing in the packageName would be more inline with what we do in other tools, and we require that rootDir end in packageName. packageName is already available in plasses calling into here.

FredKSchott: sounds good to me. My intention was not to further couple build with polyserve's specific sibling layout, but if that's already a given then that makes sense as well.

src/babel-plugin-bare-specifiers.ts
```diff
@@ -76,19 +77,22 @@ export const resolveBareSpecifiers = (
...
           !pathIsInside(filePath, componentDir)) {
+        if (packageName === undefined) {
+          throw new Error(
+              'resolveBareSpecifiers(): packageName must be set when isComponentRequest === true');
+        }
         const relativePathFromComponentDir =
             relative(componentDir, resolvedSpecifier);
-        // Important: Create a URL pointing one level up from the root dir
+        // Important: Create a URL pointingup from the root dir
+        const relativeDistance =
```
justinfagnani: This is ok for for now, but I was hoping to get to get away from detecting the `@` in the package name and do the operations all based on paths. I think a refactoring that pulls out all off these settings into some kind of callback to customize would be nice.

FredKSchott: gotcha, I didn't understand that. This is all new logic to me so I'm less familiar with it, if you can give some guidance on how that would work I can add it now.

justinfagnani: Don't worry about it for now, I think it might require a little more logic and this works, so let's lock that in. I'm not sure the value would be there now.

PR #320, opened 2018-03-21
 Compile with strict null checks.

rictic:
Also add a tsc-then test task for vscode.

This should otherwise be a noop change.


src/analyzer.ts
```diff
@@ -507,12 +514,12 @@ export class StreamLoader implements UrlLoader {
...
     }
 
-    const urlPath = urlObject.pathname;
+    const urlPath = urlObject.pathname || '/';
     const filePath = pathFromUrl(this.config.root, urlPath);
     const file = this._buildAnalyzer.getFile(filePath);
 
     if (file) {
-      return file.contents.toString();
+      return file.contents!.toString();
```
rictic: My general strategy was to assume that `file.contents` is always defined, because that's the only kind of Vinyl files we support in build.

rictic:
(also it's a way smaller PR than it looks. I updated dependencies so of course there ended up being a huge package.json change...)

package.json
```diff
@@ -8,9 +8,9 @@
     "lint": "gulp lint",
     "build": "gulp build",
     "test": "gulp lint test",
+    "test:watch": "tsc-then -- mocha -c",
     "prepublishOnly": "npm test",
-    "format": "find src/ -iname '*.ts' -o -iname '*.js' | xargs clang-format --style=file -i",
-    "test:watch": "watchy -w src/ -- npm test --loglevel=silent"
+    "format": "find src/ -iname '*.ts' -o -iname '*.js' | xargs clang-format --style=file -i"
```
stramel: you can simplify your `find` args.

PR #322, opened 2018-03-21
fix isComponentRequest check

FredKSchott:


src/babel-plugin-bare-specifiers.ts
```diff
@@ -77,7 +77,7 @@ export const resolveBareSpecifiers = (
       // If we have an import that crosses the top-level / componentDir
       // boundary, fix up the path to generate polyserve-style ../ cross-package
       // imports.
-      if (isComponentRequest !== undefined &&
+      if (isComponentRequest &&
```
justinfagnani: `isComponentRequest === true` might prevent a regression. either way

PR #327, opened 2018-03-22
Eliminate noisy test output.

rictic:
When there are no unexpected errors, test output should only contain the names of tests.

Also, ignore `not-loadable` warnings, as polymer-build should not load remote resources.


src/test/analyzer_test.ts
```diff
@@ -338,7 +338,7 @@ suite('Analyzer', () => {
       ],
       sources: ['a.html', 'b.html', 'entrypoint.html'],
     });
-    const analyzer = new BuildAnalyzer(config);
+    const analyzer = new BuildAnalyzer(config, null);
```
usergenic: Why can't we rely on default instead of these explicit nulls?

rictic: Explicit null is different. It's explicitly saying "no output please".

`undefined` is what you pass in if you want the default.

usergenic: Or to put it another way, why can't it default to null?

usergenic: OH ... sorry I see why.

usergenic:
Lets merge this before the 3.0.0-pre.1 release.

PR #328, opened 2018-03-22
Use FsUrlLoader.getFilePath instead of pathFromUrl.

rictic:
- [x] CHANGELOG.md not updated, noop change.


rictic:
Looks like the way that analyzer canonicalizes urls is causing windows troubles. Withdrawing this code review for a bit.

PR #329, opened 2018-03-22
Add htmlTransform function to build, based on Polyserve compile middleware.

aomarks:
This is based on the Polyserve compile middleware at https://github.com/Polymer/polyserve/blob/master/src/compile-middleware.ts and will replace most of that module in an upcoming PR.

It has been rewritten fairly significantly to integrate ordered script execution, which was based on Polymer/polyserve@fc33cd5

Also adds TypeScript source map to the build.

**NOTE:** I suggest looking at the `chains inline and external module scripts` test first to see if it makes sense.


src/html-transform.ts
```diff
@@ -0,0 +1,256 @@
...
+        p.NOT(p.hasAttr('type')),
+        p.hasAttrValue('type', 'text/javascript'),
+        p.hasAttrValue('type', 'application/javascript'),
+        p.hasAttrValue('type', 'module')));
+
+/**
+ * Options for htmlTransform.
+ */
+export interface HtmlTransformOptions {
+  // Transformations to apply to JavaScript within the HTML document.
```
justinfagnani: these should be `/**` comments

src/html-transform.ts
```diff
@@ -0,0 +1,256 @@
...
+  // defer running tests until our AMD modules have loaded. If WCT finds a
+  // `waitFor`, it passes it a callback that will run the tests, instead of
+  // running tests immediately.
+  //
+  // Note we must do this as late as possible, before the WCT script, because
+  // users may be setting their own `waitFor` that musn't clobber ours. Likewise
+  // we must call theirs if we find it.
+  dom5.insertBefore(wctScript.parentNode!, wctScript, parse5.parseFragment(`
+<script>
+  // Injected by Polyserve.
```
justinfagnani: expand this just a little to add a why

src/html-transform.ts
```diff
@@ -0,0 +1,256 @@
...
+</script>
+`));
+
+  // Monkey patch `require` to keep track of loaded AMD modules. Note this
+  // assumes that all modules are registered before `DOMContentLoaded`, but
+  // that's an assumption WCT normally makes anyway. Do this right after
+  // RequireJS is loaded, and hence before the first module is registered.
+  //
+  // TODO We may want to detect when the module failed to load (e.g. the deps
+  // couldn't be resolved, or the factory threw an exception) and show a nice
```
justinfagnani: file an issue

src/html-transform.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  if (wctScript && requireJsScript) {
+    addWctTimingHack(wctScript, requireJsScript);
+  }
+
+  return parse5.serialize(document);
+}
+
+function transformEsModulesToAmd(
+    moduleScripts: dom5.Node[], jsOptions: JsTransformOptions|undefined) {
```
justinfagnani: can we just pass all module + defer scripts here and get the correct total ordering?

src/html-transform.ts
```diff
@@ -0,0 +1,256 @@
...
+          script, `define('${generatedModule}', [${depsStr}]);`);
+
+    } else {
+      // Transform inline scripts with the AMD Babel plugin transformer.
+      const depStr = previousGeneratedModule === undefined ?
+          '' :
+          `'${previousGeneratedModule}', `;
+      dom5.setTextContent(
+          script,
+          jsTransform(
```
justinfagnani: style nit: I think this might read more linearly and format nicer if you put the new source in a variable

src/html-transform.ts
```diff
@@ -0,0 +1,256 @@
...
+  const document = parse5.parse(html);
+  const allScripts = dom5.queryAll(document, isJsScriptNode);
+
+  let wctScript;
+  const moduleScripts = [];
+  const otherScripts = [];
+  let shouldTransformEsModuleToAmd =
+      options.js && options.js.transformEsModulesToAmd;
+
+  for (const script of allScripts) {
```
justinfagnani: Are the two passes needed? It seems like given that we want to handle defer scripts correctly eventually, we could just loop through all the scripts, keep the state needed to enforce the ordering, and dispatch to helper methods based on type. Then we don't have 3 loops, which might be easier to read too.

src/html-transform.ts
aomarks: Done.

src/html-transform.ts
aomarks: Done.

src/html-transform.ts
aomarks: Discussed offline. We don't think we need this comment/issue.

src/html-transform.ts
aomarks: Done.

src/html-transform.ts
aomarks: Done.

src/html-transform.ts
aomarks: Added defer scripts to the module chain and added tests.

PR #330, opened 2018-03-22
Node resolution is now used for paths, not just bare specifiers.

aomarks:
Node module resolution includes resolving e.g. "./foo" to "./foo.js". We will now perform resolution on existing paths, where previously we always assumed paths did not need resolution.

Addresses https://github.com/Polymer/polymer-analyzer/issues/921 for build/polyserve.

Includes some minor test refactoring (organized into a suite), but the one to pay attention to is `paths that still need node resolution`.

### DIFF
--- a/src/babel-plugin-bare-specifiers.ts
+++ b/src/babel-plugin-bare-specifiers.ts
@@ -16,6 +16,7 @@ import {dirname, join, relative} from 'path';
 
 import resolve = require('resolve');
 import {NodePath} from 'babel-traverse';
+import * as pathlib from 'path';
 import * as isWindows from 'is-windows';
 import * as whatwgUrl from 'whatwg-url';
 import {ImportDeclaration, ExportNamedDeclaration, ExportAllDeclaration} from 'babel-types';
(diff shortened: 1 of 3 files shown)


src/babel-plugin-bare-specifiers.ts
```diff
@@ -56,29 +57,31 @@ export const resolveBareSpecifiers = (
         return;
       }
 
-      if (isPathSpecifier(specifier)) {
+      let resolvedSpecifier;
+      try {
+        resolvedSpecifier = resolve.sync(specifier, {
+          basedir: pathlib.dirname(filePath),
+          extensions: ['.js', '.json'],
```
justinfagnani: Why add `.json` support? It won't do anything good in a browser.

aomarks: Fair enough, removed.

The only reason to keep it would be so that we are exactly following the Node resolution algorithm (https://nodejs.org/api/modules.html#modules_all_together).

The difference in practice if this case came up would be that the runtime error would be "Failed to load module script: The server responded with a non-JavaScript MIME type of "application/json". Strict MIME type checking is enforced for module scripts per HTML spec." instead of a 404.

src/test/js-transform_test.ts
```diff
@@ -124,92 +124,119 @@ suite('jsTransform', () => {
...
+        import { qux } from './qux';
+      `);
 
-    const input = stripIndent(`
-      import { dep1 } from 'dep1';
-    `);
+      const expected = stripIndent(`
+        import { bar } from './bar.js';
+        import { baz } from './baz.json';
+        import { qux } from './qux';
```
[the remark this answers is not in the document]
aomarks: It's testing that if the exact extension-less file exists, it will be preferred over .js (which is what Node does). I split it into a separate test to self-document it.

src/babel-plugin-bare-specifiers.ts
aomarks: Discussed offline. We now resolve `.json` and `.node`, but we warn when you got something other than `.js`. Also added warning for when you had a bare module specifier, but it could not be resolved.

aomarks:
PTAL

PR #334, opened 2018-03-26
Transform ES to AMD modules in build streams.

### LINKED ISSUE #269: Support transpilation of ES modules (both inline and external scripts)

aomarks:
Fixes #269, but will need integration into CLI and a little more testing.

- We now do ES to AMD module transformation from `getOptimizeStreams`. This was kind of tricky because of the `HtmlSplitter` flow. Previously, transformation of inline JS did not depend on any context from the document it came from. Now, it depends on the order that the script tag appeared in the document, because we maintain ordering of `type=module` scripts by generating an artificial AMD dependency chain.

- We now inject an inline minified RequireJS AMD loader, instead of an external script. This is built by our gulpfile, exactly like the Babel helpers. This means that the `getOptimizeStreams` options now must provide the entrypoint path, in order for us to know where to inject it. (As an aside, I feel like the API/naming around "optimize streams" doesn't make a whole lot of sense anymore).

- We now inject the Babel helpers from the `htmlTransform` function and `getOptimizeStreams`, whereas we used to use a separate transform step. This is some consolidation which eliminates redundant parsing. Maybe we should do the same with the ES5 adapter? Is this making our API better (less parsing, maybe simpler to understand) or worse (less modular, maybe harder to understand)? I don't know, but there it is.

- The `htmlTransform` function now does minification, whereas it used to be a separate transformer. Again, just some consolidation.

- The `htmlTransform` function now removes the phantom `<html>`, `<head>`, and `<body>` tags injected by parse5.


aomarks:
@justinfagnani This is ready for review.

src/html-splitter.ts
```diff
@@ -148,32 +155,50 @@ class HtmlSplitTransform extends AsyncTransformStream<File, File> {
...
-          path: childPath,
-          contents: new Buffer(source),
-        });
-        this._state.addSplitPath(filePath, childPath);
-        this.push(scriptFile);
+        const isModule = typeAttribute === 'module';
+        const isInline = !dom5.hasAttribute(scriptTag, 'src');
+
+        if (isInline) {
+          const childFilename =
```
justinfagnani: consider a comment saying what we do for inline scripts

src/html-splitter.ts
```diff
@@ -148,32 +155,50 @@ class HtmlSplitTransform extends AsyncTransformStream<File, File> {
...
+          const childPath =
+              osPath.join(osPath.dirname(filePath), childFilename);
+          scriptTag.childNodes = [];
+          dom5.setAttribute(scriptTag, 'src', childFilename);
+          dom5.setAttribute(scriptTag, htmlSplitterAttribute, '');
+          const scriptFile = new File({
+            cwd: file.cwd,
+            base: file.base,
+            path: childPath,
+            contents: new Buffer(source),
```
justinfagnani: FYI, `new Buffer()` is deprecated: https://nodejs.org/api/buffer.html#buffer_new_buffer_buffer
but Buffer.from isn't in Node 4

src/html-transform.ts
```diff
@@ -102,52 +123,89 @@ export function htmlTransform(
...
+
+    // Inject as late as possible (just before the first module is declared, if
+    // there is one) because there may be some UMD dependencies that we want to
+    // continue to load in global mode instead of AMD mode (which is detected by
+    // the presence of the `require` global).
+    if (firstModuleScript !== undefined) {
       dom5.insertBefore(
           firstModuleScript.parentNode!, firstModuleScript, fragment);
+    } else {
+      // TODO(aomarks) If there were no modules, where do we put the AMD loader?
```
justinfagnani: we don't need to?

src/html-transform.ts
```diff
@@ -239,3 +290,48 @@ function addWctTimingHack(wctScript: dom5.Node, requireJsScript: dom5.Node) {
...
+}
+
+/**
+ * parse5 will inject <html>, <head>, and <body> tags if they aren't already
+ * there. Undo this so that we make fewer unnecessary transformations.
+ *
+ * Note that the given document must have been parsed with `locationInfo: true`,
+ * or else this function will always remove these tags.
+ */
+function removeFakeNodes(document: dom5.Node) {
```
justinfagnani: This should be in dom5

src/optimize-streams.ts
```diff
@@ -109,15 +108,51 @@ export class JsTransform extends GenericOptimizeTransform {
...
+      });
+    };
+
+    super('js-transform', transformer);
+  }
+}
 
-    super('babel-compile', transformer);
+/**
+ * Transform JavaScript.
```
justinfagnani: update comment

src/html-transform.ts
aomarks: I was thinking that you could be an entrypoint that has no modules of its own, but has an HTML import which uses modules...

src/html-splitter.ts
aomarks: Do you mean how Html splitter works, or how our modules transformation works?

src/html-splitter.ts
aomarks: Ok, I guess I don't want to change that in this PR though. These lines changed only because I intended them.

src/html-transform.ts
aomarks: Added TODO.

src/optimize-streams.ts

PR #338, opened 2018-03-28
Upgrade bundler to 4.0.0 pre.3

usergenic:
Bundler changed the result of buildDepsIndex which build uses, so there's a deletion of a property reference in one place but that was all the code change necessary *phew*.


usergenic:
@rictic Are you saying I make this the release PR? I usually keep code change and release PR separate as a matter of consistent practice.

rictic:
When it's a small enough change like this I think it's kosher

usergenic:
Well, at any rate this next release PR is going to help with our code-review turnaround metrics. :-p

PR #341, opened 2018-03-29
A few changes for Polyserve integration.

aomarks:
- Expose `htmlTransform` from main package index.

- AMD loader will now only be injected into an HTML document if it contains at least one `type=module` script.

- Added `softParseError` option to `jsTransform`. If set, Babel parse errors will no longer throw. Instead, a console error will be logged, and the original JS returned.

- Prepare `3.0.0-pre.5` release.

### DIFF
--- a/src/html-transform.ts
+++ b/src/html-transform.ts
@@ -133,7 +133,8 @@ export function htmlTransform(
         finalModuleScript.parentNode!, finalModuleScript, fragment);
   }
 
-  if (options.injectAmdLoader) {
+  if (options.injectAmdLoader && shouldTransformEsModuleToAmd &&
+      firstModuleScript !== undefined) {
     const fragment = parse5.parseFragment('<script></script>\n');
     dom5.setTextContent(fragment.childNodes![0], getMinifiedRequireJs());
     const requireJsScript = fragment.childNodes![0];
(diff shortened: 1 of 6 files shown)


src/js-transform.ts
```diff
@@ -156,7 +161,19 @@ export function jsTransform(js: string, options: JsTransformOptions): string {
   }
 
   if (doBabel) {
-    js = babelCore.transform(js, {presets, plugins}).code!;
+    try {
+      js = babelCore.transform(js, {presets, plugins}).code!;
+    } catch (e) {
```
rictic: This will catch more than just parse errors. Maybe we should inspect the errors?

src/html-transform.ts
```diff
@@ -133,7 +133,7 @@ export function htmlTransform(
         finalModuleScript.parentNode!, finalModuleScript, fragment);
   }
 
-  if (options.injectAmdLoader) {
+  if (options.injectAmdLoader && firstModuleScript !== undefined) {
```
rictic: This is subtle and easy to see it breaking. Would a quick test be warranted?

aomarks: Done.

aomarks:
PTAL

src/js-transform.ts
aomarks: Done. Is this too brittle? Also renamed to `softSyntaxError` to match Babel's terminology.

rictic: What do you think about the (roughly equivalent): `e.constructor.name === 'SyntaxError'`

src/test/html-transform_test.ts
```diff
@@ -297,6 +297,22 @@ suite('htmlTransform', () => {
...
+          <script src="depA.js"></script>
+        </body></html>`;
+
+      const result = htmlTransform(input, {
+        injectAmdLoader: true,
+        js: {
+          transformEsModulesToAmd: true,
+        },
+      });
+      assertEqualIgnoringWhitespace(replaceGiantScripts(result), input);
```
rictic: Do we need to call `replaceGiantScripts` here?

aomarks: Done.

PR #342, opened 2018-03-29
Update some dependencies

justinfagnani:


justinfagnani:
Oops, this includes a commit from #336. Wait until I commit that.

justinfagnani:
@rictic that's a bit of a change because we need to change a bunch of callsites and bring in IterableX or something to replace the array methods. This was me quickly updating some deps so we'd have a slightly smaller npm installation.

package.json
```diff
@@ -30,11 +30,11 @@
     "@types/is-windows": "^0.2.0",
     "@types/mz": "0.0.31",
     "@types/node": "^6.0.77",
-    "@types/parse5": "^2.2.34",
+    "@types/parse5": "^3.0.0",
```
rictic: This is a bad version, it's a stub with no typings, so this does not compile (parse5 v3 added its own typings, but they were bad, so parse5 v4 dropped them again, kinda). Sending over a fix.

PR #345, opened 2018-03-30
Integrate AMD + module bundling

aomarks:
A few fixes to get ES module bundling integrated into polymer-build.

- Write out generated JS bundles. Previously we were receiving parse5 ASTs and calling serialize on them, which produced nothing for JS bundles. Now we receive the JS or or HTML source directly.
- Delete JS files that have been placed into a bundle.
- Pass `moduleResolution` down to the build analyzer.
- Export the `matchesExt` function, CLI will need it in an upcoming PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5892,9 +5892,9 @@
       }
     },
     "polymer-bundler": {
-      "version": "4.0.0-pre.3",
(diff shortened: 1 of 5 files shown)


aomarks:
Sorry, more fixes to make here, hold off on review, will re-add later.

PR #346, opened 2018-03-30
Upgrade to Babel 7

justinfagnani:


gulpfile.js
```diff
@@ -44,8 +44,8 @@ gulp.task('build', (done) => {
 });
 
 gulp.task('compile', () => {
-  const srcs = gulp.src(['src/**/*.ts', 'custom_typings/**/*.ts']);
-  const tsResult = srcs.pipe(sourcemaps.init())
+  const tsResult = gulp.src(['src/**/*.ts', 'custom_typings/**/*.ts'])
```
rictic: Arg, we should really be getting sources from the the tsproject

(ok really really we should be dropping gulp)

package.json
```diff
@@ -35,22 +46,7 @@
...
-    "babel-plugin-syntax-object-rest-spread": "^6.13.0",
-    "babel-plugin-transform-async-generator-functions": "^6.24.1",
-    "babel-plugin-transform-async-to-generator": "^6.24.1",
-    "babel-plugin-transform-es2015-modules-amd": "^6.24.1",
-    "babel-plugin-transform-exponentiation-operator": "^6.24.1",
-    "babel-plugin-transform-object-rest-spread": "^6.26.0",
-    "babel-preset-es2015": "^6.24.1",
-    "babel-preset-minify": "^0.3.0",
-    "babel-traverse": "^6.26.0",
+    "babel-preset-minify": "^0.4.0-alpha.caaefb4c",
```
rictic: I think we should pin to that version of preset-minify, as ^ isn't likely to do what we want when ordering SHAs like `caaefb4c`

src/js-transform.ts
```diff
@@ -24,43 +24,39 @@ import {resolveBareSpecifiers} from './babel-plugin-bare-specifiers';
...
 // Syntax and transform plugins for ES2015.
 const babelPresetEs2015NoModules =
-    require('babel-preset-es2015').buildPreset({}, {modules: false});
+    [require('@babel/preset-es2015'), {modules: false}];
 
 // The ES2016 and ES2017 presets do not inherit the plugins of previous years,
 // and there is no ES2018 preset yet. Since the additions in ES2016 and ES2017
 // are small, and we have to list syntax plugins separately anyway (see below),
 // just enumerate the transform plugins here instead of merging the 3 presets.
 const babelTransformPlugins = [
```
rictic: TODO: use babel's env instead?

aomarks: There's a TODO right above already.

gulpfile.js
justinfagnani: You somehow picked up a commit I removed before making the PR. Weird... This commit's already merged.

package.json
justinfagnani: done

PR #347, opened 2018-03-30
Implement import.meta transform

justinfagnani:

### DIFF
--- a/src/babel-plugin-import-meta.ts
+++ b/src/babel-plugin-import-meta.ts
@@ -13,19 +13,31 @@
  */
 
 import importMetaSyntax from '@babel/plugin-syntax-import-meta';
+import template from '@babel/template';
 import {NodePath} from '@babel/traverse';
-import * as t from '@babel/types';
 import {MetaProperty} from '@babel/types';
 import {relative} from 'path';
-import {URL} from 'whatwg-url';
+
+const ast = template.ast;
 
 /**
  * Rewrites `import.meta` into an object with a `url` property.
+ *
+ * `import.meta.url` must be a URL string with the fully qualified URL of the
+ * module. We use the document's base URI and the relative path from rootDir to
+ * filePath to build the URL.
+ *
+ * TODO: Add a option for a path fragment between the document base and module
+ * file path.
+ *
+ * @param filePath THe path of the file being transformed
+ * @param rootDir The root project folder containing filePath
+ * @param base A base URL to use instead of document.baseURI
  */
 export const rewriteImportMeta = (
     filePath: string,
     rootDir: string,
-    base: string,
+    base?: string,
     ) => ({
   inherits: importMetaSyntax,
   manipulateOptions(_opts: any, parserOpts: any) {
@@ -36,12 +48,10 @@ export const rewriteImportMeta = (
       const node = path.node;
       if (node.meta && node.meta.name === 'import' &&
           node.property.name === 'meta') {
-        // console.log('import.meta at', node.loc);
         const relativePath = relative(rootDir, filePath);
-        const url = new URL(relativePath, base);
+        const baseURI = base !== undefined ? `'${base}'` : 'document.baseURI';
         path.replaceWith(
-            t.objectExpression([t.objectProperty(
-                t.identifier('url'), t.stringLiteral(url.toString()))]) as any);
+            ast`({url: new URL('${relativePath}', ${baseURI}).toString()})`);
       }
     }
   }
--- a/src/babel-plugin-import-meta.ts
+++ b/src/babel-plugin-import-meta.ts
@@ -21,14 +21,14 @@ import {relative} from 'path';
 const ast = template.ast;
 
 /**
- * Rewrites `import.meta` into an object with a `url` property.
+ * Rewrites `import.meta`[1] into an object with a `url`[2] property.
  *
  * `import.meta.url` must be a URL string with the fully qualified URL of the
  * module. We use the document's base URI and the relative path from rootDir to
  * filePath to build the URL.
  *
- * TODO: Add a option for a path fragment between the document base and module
- * file path.
+ * [1]: https://github.com/tc39/proposal-import-meta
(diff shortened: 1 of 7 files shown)


src/babel-plugin-import-meta.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import importMetaSyntax from '@babel/plugin-syntax-import-meta';
+import {NodePath} from '@babel/traverse';
+import * as t from '@babel/types';
+import {MetaProperty} from '@babel/types';
+import {relative} from 'path';
+import {URL} from 'whatwg-url';
+
+/**
+ * Rewrites `import.meta` into an object with a `url` property.
```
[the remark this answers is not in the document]
justinfagnani: done

justinfagnani:
PTAL

src/babel-plugin-import-meta.ts
```diff
@@ -0,0 +1,57 @@
...
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import importMetaSyntax from '@babel/plugin-syntax-import-meta';
+import template from '@babel/template';
```
aomarks: `import {ast} from '@babel/template'`?

src/babel-plugin-import-meta.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Rewrites `import.meta`[1] into an object with a `url`[2] property.
+ *
+ * `import.meta.url` must be a URL string with the fully qualified URL of the
+ * module. We use the document's base URI and the relative path from rootDir to
+ * filePath to build the URL.
+ *
+ * [1]: https://github.com/tc39/proposal-import-meta
+ * [2]: https://html.spec.whatwg.org/#hostgetimportmetaproperties
+ *
+ * @param filePath THe path of the file being transformed
```
aomarks: s/THe/The

src/babel-plugin-import-meta.ts
```diff
@@ -0,0 +1,57 @@
...
+    inherits: importMetaSyntax,
+    visitor: {
+      MetaProperty(path: NodePath<MetaProperty>) {
+        const node = path.node;
+        if (node.meta && node.meta.name === 'import' &&
+            node.property.name === 'meta') {
+          const relativePath = relative(rootDir, filePath);
+          const baseURI = base !== undefined ? `'${base}'` : 'document.baseURI';
+          path.replaceWith(
+              ast`({url: new URL('${relativePath}', ${baseURI}).toString()})`);
```
aomarks: `ast` is a cool template literal tag

src/babel-plugin-import-meta.ts
```diff
@@ -0,0 +1,57 @@
...
+    base?: string,
+    ) => {
+  return {
+    inherits: importMetaSyntax,
+    visitor: {
+      MetaProperty(path: NodePath<MetaProperty>) {
+        const node = path.node;
+        if (node.meta && node.meta.name === 'import' &&
+            node.property.name === 'meta') {
+          const relativePath = relative(rootDir, filePath);
```
aomarks: If either `rootDir` or `filePath` are relative, `path.relative` will first evaluate them as relative to `cwd`, and then compute a relative path from one to the other. I feel like this could be a little confusing.

Maybe another approach would be to just take the final relative URL, to shift the burden of computing a relative path (if needed) to caller?

Also on windows `relative` will return a backslash delimited path, so that would need to be fixed either way I think.

src/test/js-transform_test.ts
```diff
@@ -280,4 +280,28 @@ suite('jsTransform', () => {
     const result = jsTransform(input, {transformEsModulesToAmd: true});
     assert.equal(result.trim(), expected.trim());
   });
+
+  test('transforms import.meta', () => {
+    const rootDir =
+        path.join(__dirname, '..', '..', 'test-fixtures', 'npm-modules');
+    const filePath = path.join(rootDir, 'npm-module.js') as LocalFsPath;
```
aomarks: Don't actually need the `LocalFsPath` cast (optionally update the other cases too).

src/babel-plugin-import-meta.ts
justinfagnani: > Maybe another approach would be to just take the final relative URL, to shift the burden of computing a relative path (if needed) to caller?

I waved on this myself. I thought that if the computation of the final relative path was always the same, then we could safer handle it here. Your points swayed me though. It's also possible that a caller just already has the relative path.

With this change I think it should be a relative URL though, and not do any Windows handling here.

src/babel-plugin-import-meta.ts
justinfagnani: Yep. I'm ready to use this everywhere else now.

src/babel-plugin-import-meta.ts
justinfagnani: unfortunately, it's a default export of an object with an `ast` property.

src/test/js-transform_test.ts
justinfagnani: done

src/js-transform.ts
```diff
@@ -151,6 +158,21 @@ export function jsTransform(js: string, options: JsTransformOptions): string {
...
   }
+  if (options.transformImportMeta) {
+    if (!options.filePath) {
+      throw new Error('Cannot perform importMeta transform without filePath.');
+    }
+    if (!options.rootDir) {
+      throw new Error('Cannot perform importMeta transform without rootDir.');
+    }
+    doBabel = true;
+    let relativeURL = relative(options.rootDir, options.filePath);
```
aomarks: Can you update the description for filepath and rootdir to say that they must both be absolute system paths?

justinfagnani: done

PR #350, opened 2018-04-03
Transform module

aomarks:


src/optimize-streams.ts
```diff
@@ -102,19 +104,23 @@ export class GenericOptimizeTransform extends Transform {
  * Transform JavaScript.
  */
 export class JsTransform extends GenericOptimizeTransform {
-  constructor(options: JsOptimizeOptions) {
+  constructor(options: OptimizeOptions) {
+    const jsOptions = options.js || {};
```
justinfagnani: 🤡🏌️: `const {jsOptions = {}} = options;`

aomarks: I think that would be `const {js: jsOptions = {}} = options;` which is slightly longer and probably more confusing than the original?

